PR #10, opened 2021-08-30
Correct bounding box calculation for drilled slot.

eyal0:
The conversion of units was done in the wrong place leading to the bounding box to be incorrect.  This fixes it.

Also, an em-dash was used instead of a minus character in the file, causing the file to not be ASCII.

### DIFF
--- a/src/drill.c
+++ b/src/drill.c
@@ -480,6 +480,12 @@ parse_drillfile(gerb_file_t *fd, gerbv_HID_Attribute *attr_list, int n_attr, int
 		curr_net->stop_x = state->curr_x;
 		curr_net->stop_y = state->curr_y;
 
+		if (state->unit == GERBV_UNIT_MM) {
+		    /* Convert to inches -- internal units */
+		    curr_net->stop_x /= 25.4;
+		    curr_net->stop_y /= 25.4;
+		}
+
 		r = image->aperture[state->current_tool]->parameter[0]/2;
 
 		/* Update boundingBox with drilled slot stop_x,y coords */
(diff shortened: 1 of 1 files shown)


ooxi:
Thank you very much for your pull request @eyal0! I have already merged the ASCII fix in 17b9cb6389cfea394a56f3e8e441e69d6cea5b01.

I have already seen your patch regarding the wrong calculation and I think you are right. Nevertheless, could you provide a test file where I can verify the problem? That would be great :-)

eyal0:
Sure, no problem!  It's good that you ask!  Here's a file that you can try:

```
M48
;DRILL file {KiCad 4.0.4+e1-6308~48~ubuntu14.04.1-stable} date Thu Nov 10 21:17:09 2016
;FORMAT={-:-/ absolute / metric / decimal}
FMAT,2
METRIC,TZ
T1C0.400
T2C1.001
T3C3.000
T4C4.000
%
G90
G05
M71
T1
X86.36Y-72.39
X86.36Y-80.55
X86.36Y-82.55
X86.36Y-84.55
X88.86Y-72.39
X113.07Y-72.39
X114.3Y-80.55
X114.3Y-82.55
X114.3Y-84.55
X115.57Y-72.39
T2
X82.55Y-64.77
X82.55Y-74.93
X92.71Y-64.77
X92.71Y-74.93
X95.25Y-83.82
X97.79Y-83.82
X102.87Y-83.82
X105.41Y-83.82
X109.22Y-64.77
X109.22Y-74.93
X119.38Y-64.77
X119.38Y-74.93
T2
X96.901Y-65.701G85X96.901Y-67.7
G05
X102.599Y-63.701G85X100.601Y-63.701
G05
T3
X102.599Y-69.7G85X100.601Y-60.7
G05
T0
M30
```

When I open it in `gerbv` with the patch installed, I see this:

good result

Without the patch:

bad result

Notice how the ruler around the image is different.  This is because gerbv has incorrectly calculated the correct bounding box.  This error also shows up in the bounding box that is reported by `libgerbv` if you use it as a library without the viewer.

I hope that makes sense!  Let me know if you need more info.

PR #12, opened 2021-08-30
Set the aperature earlier

eyal0:
If we never enter the loop on line 1128 then the test on line 1231 is examining a variable which has never been initialized.  This is undefined behavior and was caught by valgrind.

This fixes the following valgrind error:

```
==44371== Conditional jump or move depends on uninitialised value(s)
==44371==    at 0x4053E9A: drill_parse_T_code (drill.c:1232)
==44371==    by 0x4052F17: parse_drillfile (drill.c:782)
==44371==    by 0x4066BCB: gerbv_open_image (gerbv.c:532)
==44371==    by 0x4065E4F: gerbv_open_layer_from_filename (gerbv.c:230)
==44371==    by 0x210C71: GerberImporter::load_file(std::__cxx11::basic_string<char, std::char_traits<char>
==44371==    by 0x3E4E23: do_pcb2gcode(int, char const**) (main.cpp:343)
==44371==    by 0x1DDF22: main (main.cpp:409)
```

### DIFF
--- a/src/drill.c
+++ b/src/drill.c
@@ -1119,6 +1119,7 @@ drill_parse_T_code(gerb_file_t *fd, drill_state_t *state,
 
     /* Set the current tool to the correct one */
     state->current_tool = tool_num;
+    apert = image->aperture[tool_num];
 
     /* Check for a size definition */
     temp = gerb_fgetc(fd);
@@ -1155,7 +1156,6 @@ drill_parse_T_code(gerb_file_t *fd, drill_state_t *state,
 			    "at line %ld in file \"%s\""),
 			    size, tool_num, file_line, fd->filename);
 	    } else {
-		apert = image->aperture[tool_num];
 		if (apert != NULL) {
 		    /* allow a redefine of a tool only if the new definition is exactly the same.
 		     * This avoid lots of spurious complaints with the output of some cad


ooxi:
Thank you very much for your PR @eyal0!

Could you describe why we set `apert` to `image->aperture[tool_num];` instead of `NULL`? Unfortunately I'm not overly familiar with the parsing code but it seems like we should set `apert` only for specific conditions (like `temp` was `'C'`) and have cleanup code if that condition was never reached.

eyal0:
So that `apert` is always a cached copy of `image->aperture[tool_num]`.  It makes it easier to insert more code into that function.  For example, if later on someone wants to put some code before the for loop that looks at `apert` and tries to do something with it then they'll again need to fix this.  With this patch, `apert` is available and in sync with `tool_num` as soon as possible.

I think that it's a nicer solution than just setting to `NULL` to dodge a valgrind error.  Setting to `NULL` to dodge a valgrind error just hides bad programming instead of fixing bad programming!

I think that the result would be the same either way.  Up to you!

eyal0:
Also, if you set to `NULL`, you can't delete line 1158.  Even worse, if someone wants to implement the cases on line 2015, not only can you not delete line 1158, you will need to copy line 1158 to line 2015.  Potentially you will need to copy that line 1158 to all the cases!  Better to just do it once at the top.

ooxi:
Since you are much more familiar with the source code, I'm inclined to trust your judgement :-)

Nevertheless, shouldn't we also delete [Line 1229-1259](https://github.com/gerbv/gerbv/blob/main/src/drill.c#L1229-L1259) if we set `apert` to `image->aperture[tool_num]` since it cannot be `NULL` now? Or is it possible for `image->aperture[tool_num]` to be `NULL`, too?

ooxi:
Thank you for taking your time @eyal0, I didn't grasp the logic of `drill_parse_T_code` at first! Merged as `425d48da0bdb0de3d576658afec21d0f523ae9dd`.

eyal0:
It can still be null, as you noticed. The tools start as an array of NULLs and as they are defined, they are malloc'd.

ooxi:
Thank you very much @eyal0 for providing those excellent steps to reproduce the issue. Your patch was merged into master.

As discussed via mail, I'm very glad that you are willing to act as comaintainer and help out with pull request analysis.

PR #17, opened 2021-09-03
ci: Build and test

### LINKED ISSUE #16: Add run_tests.sh to the CI
We should be running tests in CI after the build.

This issue can't be fixed until #15 is fixed.

eyal0:
Build and test the code as part of CI.

This fixes #16

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -17,6 +17,23 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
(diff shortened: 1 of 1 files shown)


ooxi:
How about we run the tests, but do not fail the build? There seems to be [continue-on-error](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error) but as far as I understand, it's only available on the job level, not the step level.

PR #20, opened 2021-09-13
apply of '#81 Fix casting pointer to different size integer'

efa:
apply of '#81 Fix casting pointer to different size integer'
https://sourceforge.net/p/gerbv/patches/81/
as opening a .gvp file on Windows always crash

This and the next one are necessary to work on Windows too.
Follow the PR for #77 Fix double-freeing memory

### DIFF
--- a/src/scheme.c
+++ b/src/scheme.c
@@ -568,7 +568,7 @@ static int alloc_cellseg(scheme *sc, int n) {
      char *cp;
      long i;
      int k;
-     unsigned int adj=ADJ;
+     uintptr_t adj=ADJ;
 
      if(adj<sizeof(struct cell)) {
        adj=sizeof(struct cell);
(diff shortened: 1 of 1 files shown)


efa:
'size_t' has the same size of 'uintptr_t' on all 32 and 64-bit systems I know, and it is still unsigned, but it is certainly better suite for a size like adj=sizeof()

ooxi:
Thank you very much @efa for providing this patch! I have reintegrated your commit 71c623c158aae3ef45b6698032314037c277c3a5 as 26c90bd6fe153c09cb32fdb50ec1ca2078a22151 (added `stddef.h` include) in pull request #23

eyal0:
Add a static_assert that the sizeof one is equal to the sizeof the other and then you will never have a question about whether or not the size is right. The compiler will tell you the answer with no effort.

It's nice when the computer can enforce rules for us.

PR #21, opened 2021-09-13
Double free

efa:
apply of '#77 Fix double-freeing memory', only the part that was not already applyed:
https://sourceforge.net/p/gerbv/patches/77/
as opening 2nd or 3th .gvp file on Windows crash

### DIFF
--- a/src/callbacks.c
+++ b/src/callbacks.c
@@ -199,7 +199,7 @@ void open_project(char *project_filename)
 
 	/* Update the last folder */
 	g_free (mainProject->path);
-	mainProject->path = project_filename;
+	mainProject->path = g_strdup(project_filename);
 
 	gerbv_unload_all_layers (mainProject);
 	main_open_project_from_filename (mainProject, project_filename);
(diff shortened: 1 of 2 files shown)


ooxi:
adf73ec56007197b27f6143e72a13978156212bb looks very promising! I don't quite understand how that worked before, since C doesn't support default arguments. What was used as free function before?

Was this sufficient to get loading projects working on windows for you? :-)

efa:
> I don't quite understand how that worked before, since C doesn't support default arguments. What was used as free function before?

I think here that was some GTK interface that was deprecated/changed, and source was not updated till now.

> Was this sufficient to get loading projects working on windows for you? :-)

this patch solve the loading of 2nd and next gvp files.
To load the first .gvp file need the https://github.com/gerbv/gerbv/pull/20

ooxi:
Thank you very much for testing this @efa! I reintegrated your commit adf73ec56007197b27f6143e72a13978156212bb as a96b46c7249e97e950d860790b84bcdba2368f57 in pull request #24

PR #25, opened 2021-09-17
static_assert that casting a pointer to size_t is valid

eyal0:


ooxi:
Thank you @eyal0, this should help against reintroducing a bug similar to [#81: Fix casting pointer to different size integer](https://sourceforge.net/p/gerbv/patches/81/) in the future!

PR #28, opened 2021-09-30
Update supported platforms

ooxi:
The README.md very prominently lists platforms where gerbv once was build and tested. However those platforms are not currently covered by CI or manually verified.

If there is no general objection I will go ahead and reflect this in the public documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,10 +48,10 @@ GitHub organization.
 
 Gerbv has been built and tested on
 
-* Linux (from 2.2)
-* NetBSD/i386 (1.4.1)
-* NetBSD/Alpha (1.5.1)
-* Solaris (5.7 and 5.8)
+* Debian 10 (amd64)
+* Fedora 34 (amd64)
+* Ubuntu 20.04 (amd64)
+* Windows 10 (amd64 cross compiled from Fedora as well as native x86/amd64 using MSYS2)
 
 
 ## Information for developers


efa:
add Debian 10 64 bit
for Win10 specifyù:
  Win10/MSYS2 build chain
  and maybe cross-compilation on Linux/distro

efa:
the last release on Sourceforge was released for some other platforms?

efa:
we should support at least macOS 64 bit

ooxi:
> the last release on Sourceforge was released for some other platforms?
> we should support at least macOS 64 bit

I currently have no way to verify macOS. However I think GitHub Actions can use macOS as virtual environment. Do you have any experience with this?

ooxi:
> add Debian 10 64 bit
> for Win10 specify:
> Win10 64 bit cross-compilation on Linux/distro xx bit
> Win10/MSYS2 32 and 64 bit build chain

I have added Debian 10 and specified the toolchains for Windows 10 in [291962](https://github.com/gerbv/gerbv/pull/28/commits/291962935266f84a49d1e436385fc2b4f04f5adb)

efa:
for MSYS2, you put MSYS32.
MSYS2 supply both MinGW32 and MinGW64 toolchain, I normally generate Gerbv for both.

efa:
> Do you have any experience with this?

yes, some. I have cross-compilation env installed on my Debian targeting darwin18, and I use it to generate some QT and GTK app from my repos. QT is easy as many built tool are natively cross-platform, for GTK is harder.
Pre-built GTK are either:
1) shared object (.dylib on macOS) only
2) shared object and dev library but packaged to be installed on a real macOS (like Homebrew or other GNU distro for macOS)
3) shared object and dev library working with cross-compiler on Linux, but targeting X and not Quartz.
X target is old and require macOS users to install XQuartz, not a thing they want to.
GTK Quartz target in 2021 is working well, and supported, but NOT for cross-compilation, only native macOS.
So I had to cross-compile the whole GTK stack myself targeting Quarz, and was cumbersome, but now I had a version that work.
More I have a macOS virtual machine to test built binaries and .dmg

ooxi:
> for MSYS2, you put MSYS32.
> MSYS2 supply both MinGW32 and MinGW64 toolchain, I normally generate Gerbv for both.

Thank you for catching this! I'm unfamiliar with MSYS and did not know the nomenclature. Corrected this in commit 9c60da0af79a5128976e8aa338cb1317d659bb24.


> not on "GitHub Actions" and macOS, but yes some on macOS. I have cross-compilation env installed on my Debian targeting darwin18, and I use it to generate some QT and GTK app from my repos. QT is easy as many build tool are natively cross-platform, for GTK is harder.
>
> So I had to cross-compile the whole GTK stack (and all pre deps) myself targeting Quarz, and was cumbersome, but now I had a version that work.
>
> More I have a macOS virtual machine to test built binaries and .dmg

That's a really valuable asset to have! Let me toy around with GitHub Actions and see if I get gerbv compiled on native macOS so you could test it.

I propose we go ahead and merge the update without macOS and include it as soon as we have verified the binary. Would this be OK with you?

efa:
yes, generating macOS binaries for sure require times and many tests

PR #50, opened 2021-11-01
Use g_build_filename instead of strncpy to avoid warning.

### LINKED ISSUE #49: Replace useless strncat with strcpy
We already know the length in https://github.com/gerbv/gerbv/blob/b5f1eacd798f327ab319af939f89031db4b7c10a/src/gerb_file.c#L323 so just strcpy directly.

eyal0:
The strncat is pointless because we already know the size so we can
just use strcpy.  strcat would work, too, but then the code would be
taking the length of the string once again and that is the [Schlemiel Painter algorithm](https://en.wikichip.org/wiki/schlemiel_the_painter%27s_algorithm).

This fixes #49

### DIFF
--- a/src/gerb_file.c
+++ b/src/gerb_file.c
@@ -314,13 +314,9 @@ gerb_find_file(char const * filename, char **paths)
 	  /*
 	   * Build complete path (inc. filename) and check if file exists.
 	   */
-	  complete_path = (char *)g_malloc(strlen(curr_path) + strlen(filename) + 2);
+	  complete_path = g_build_filename(curr_path, filename, NULL);
 	  if (complete_path == NULL)
 	    return NULL;
-	  strcpy(complete_path, curr_path);
-	  complete_path[strlen(curr_path)] = G_DIR_SEPARATOR;
-	  complete_path[strlen(curr_path) + 1] = '\0';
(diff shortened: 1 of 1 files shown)


ooxi:
While I agree that the original way of creating the string was unnecessary, I'm really not a huge fan of primitive string manipulation, especially for code which is not on the hot path. For example I had to look up the manual of `strcpy` to remind myself that `strcpy` does indeed also copy the null terminator.

What do you think about swapping those statements with `g_snprintf`?

eyal0:
I will try another way using `g_build_filename`, which is a much better fit.

I put a printf after the filename construction and I have confirmed that the output is unchanged.  It looks like this both before and after when I run `gerbv -p jj.gvp`.  You can make `jj.gvp` by opening multiple files in gerbv and saving a new project.
```bash
└──> ./src/gerbv -p jj.gvp 
/home/esoha/private/gerbv/src/.libs/../share/gerbv/scheme/init.scm
/usr/local/share/gerbv/scheme/init.scm
/home/esoha/private/gerbv/src/.libs/init.scm
./init.scm
```

ooxi (approved):
This is _so_ much cleaner!

PR #51, opened 2021-11-02
Link with math library, which is necessary for sin, cos, etc.

eyal0:
The linker was complaining without this.


ooxi:
Thank you @eyal0, I didn't even see that warning because of the flood of other warnings.

eyal0:
Once we enable "-Werror" in the build, it'll be easier to see!

PR #52, opened 2021-11-02
Build with all warnings treated as errors

eyal0:


eyal0:
This will probably not pass until #51 and #50 are merged

ooxi:
There are a couple of more warnings stopping us from merging this. Most of them seem to be related to deprecated types like `GTypeDebugFlags` and `GTimeVal`.

I'm not sure how much refactoring is required to use the recommended types.

ooxi:
I am 100% on board with enabling `-Werror` for development and CI builds. Nevertheless I know that distributions do not want `-Werror` or at least want an easy way to turn it off.

We should therefore look at other projects for inspiration on how to provide this.

eyal0:
> There are a couple of more warnings stopping us from merging this. Most of them seem to be related to deprecated types like GTypeDebugFlags and GTimeVal.

>I'm not sure how much refactoring is required to use the recommended types.

Yup, but those errors aren't in gerbv, they are in glib itself!  glib is calling its own deprecated functions!

It would be too much for us to fix glib but we can teach the compiler to ignore errors inside of included files.  That is, if gerbv code causes the warning, we want the error, but if gerbv includes glib and glib causes the warning, we want to ignore it.

The trick for this is to make sure that glib is included not with `-I` but `-isystem`.  pcb2gcode does this.

https://stackoverflow.com/a/2590764/4454

eyal0:
We can at least merge #55 and this PR is proof that it solves the problem.  Next, we need to find a way to use Werror only when wanted.  I have done this before using a command-line flag, which we would put into the CI build.

eyal0:
I think that this should pass after #55 is in. @ooxi

ooxi:
I'll rerun the CI, if it work's I'll merge :-)

ooxi:
```
/usr/include/gtk-2.0/gtk/gtktypeutils.h:236:1: error: ‘GTypeDebugFlags’ is deprecated [-Werror=deprecated-declarations]
  236 | void            gtk_type_init   (GTypeDebugFlags    debug_flags);
      | ^~~~
In file included from /usr/include/gtk-2.0/gtk/gtkobject.h:37,
                 from /usr/include/gtk-2.0/gtk/gtkwidget.h:36,
                 from /usr/include/gtk-2.0/gtk/gtkcontainer.h:35,
                 from /usr/include/gtk-2.0/gtk/gtkbin.h:35,
                 from /usr/include/gtk-2.0/gtk/gtkwindow.h:36,
                 from /usr/include/gtk-2.0/gtk/gtkdialog.h:35,
                 from /usr/include/gtk-2.0/gtk/gtkaboutdialog.h:32,
                 from /usr/include/gtk-2.0/gtk/gtk.h:33,
                 from attribute.c:47:
```

@eyal0 this still errors out :-( is maybe a rebase necassary?

eyal0:
> maybe a rebase necassary?

That's what I think.  Let's see.

eyal0:
image

Looks okay.  If you look in there carefully, you can see `-Wall` and `-Werror` are arguments to the build.

ooxi:
Great work!

PR #53, opened 2021-11-02
Check that gerbv exists before copying it.

eyal0:
Without this, running the tests was creating an empty gerbv script and
causing failures.


ooxi (approved):
This change looks reasonable

PR #55, opened 2021-11-02
Use -isystem for gtk include

### LINKED ISSUE #54: Include system headers with isystem
> There are a couple of more warnings stopping us from merging this. Most of them seem to be related to deprecated types like GTypeDebugFlags and GTimeVal.

>I'm not sure how much refactoring is required to use the recommended types.

Yup, but those errors aren't in gerbv, they are in glib itself!  glib is calling its own deprecated functions!

It would be too much for us to fix glib but we can teach the compiler to ignore errors inside of included files.  That is, if gerbv code causes the warning, we want the error, but if gerbv includes glib and glib causes the warning, we want to ignore it.

The trick for this is to make sure that glib is included not with `-I` but `-isystem`.  pcb2gcode does this.

https://stackoverflow.com/a/2590764/4454

_Originally posted by @eyal0 in https://github.com/gerbv/gerbv/issues/52#issuecomment-957948804_

eyal0:
GTK has some warnings internally that we'd like to ignore.  This will
ignore warnings that are about code in GTK.  Warnings about code in
gerbv will not be suppressed.

This fixes #54

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -392,11 +392,12 @@ AC_CHECK_FUNCS(strlwr)
 AC_CHECK_FUNCS(realpath canonicalize_file_name)
 libiberty_NEED_DECLARATION(canonicalize_file_name)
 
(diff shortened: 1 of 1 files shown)


ooxi (approved):

ooxi:
Thank you!

Issue #57, opened 2021-11-03
Gerbv RS-274X aperture macro outline primitive integer overflow vulnerability (TALOS-2021-1405)


CiscoTalos:
TALOS-2021-1405   
CVE-2021-40394   

Gerbv RS-274X aperture macro outline primitive integer overflow vulnerability

### Summary

An integer overflow vulnerability exists in the RS-274X aperture macro outline primitive functionality of Gerbv 2.7.0 and dev (commit b5f1eacd). A specially-crafted gerber file can lead to code execution. An attacker can provide a malicious file to trigger this vulnerability.

### Tested Versions

Gerbv 2.7.0   
Gerbv dev  (commit b5f1eacd)   
Gerbv forked dev (commit 71493260)   

### Product URLs

[https://sourceforge.net/projects/gerbv/](https://sourceforge.net/projects/gerbv/)

### CVSSv3 Score

10.0 - CVSS:3.0/AV:N/AC:L/PR:N/UI:N/S:C/C:H/I:H/A:H

### CWE

CWE-190 - Integer Overflow or Wraparound

### Details

Gerbv is an open-source software that allows to view RS-274X Gerber files, Excellon drill files and pick-n-place files. These file formats are used in industry to describe the layers of a printed circuit board and are a core part of the manufacturing process.

Some PCB (printed circuit board) manufacturers use software like Gerbv in their web interfaces as a tool to convert Gerber (or other supported) files into images. Users can upload gerber files to the manufacturer website, which are converted to an image to be displayed in the browser, so that users can verify that what has been uploaded matches their expectations. Gerbv can do such conversions using the `-x` switch (export). For this reason, we consider this software as reachable via network without user interaction or privilege requirements.

Gerbv uses the function `gerbv_open_image` to open files. In this advisory we're interested in the RS-274X file-type.

    int
    gerbv_open_image(gerbv_project_t *gerbvProject, char *filename, int idx, int reload,
                    gerbv_HID_Attribute *fattr, int n_fattr, gboolean forceLoadFile)
    {
        ...        
        dprintf("In open_image, about to try opening filename = %s\n", filename);
        
        fd = gerb_fopen(filename);
        if (fd == NULL) {
            GERB_COMPILE_ERROR(_("Trying to open \"%s\": %s"),
                            filename, strerror(errno));
            return -1;
        }
        ...
        if (gerber_is_rs274x_p(fd, &foundBinary)) {                                 // [1]
            dprintf("Found RS-274X file\n");
[...]

PR #61, opened 2021-11-04
Fix CVE-2021-40394

ooxi:
A possible signed integer overflow was discovered by the @CiscoTalos team in issue #57. This patch mitigates the issue by clamping the computation.

### DIFF
--- a/src/gerber.c
+++ b/src/gerber.c
@@ -30,6 +30,7 @@
 
 #include <stdlib.h>
 #include <string.h>
+#include <limits.h>
 #include <math.h>  /* pow() */
 #include <errno.h>
 #include <ctype.h>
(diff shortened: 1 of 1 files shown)


ooxi:
This patch should also include a test to reproduce the issue as pioneered by 672214abb47a802fc000125996e6e0a46c623a4e

eyal0:
> This patch should also include a test to reproduce the issue as pioneered by [672214a](https://github.com/gerbv/gerbv/commit/672214abb47a802fc000125996e6e0a46c623a4e)

Yeah, would be nice to have it in CI, too.  I tried to build with asan but it didn't work so easy as just adding a flag like the #57 indicated.  I'll try again when I have time.

I already have Valgrind support in there.  Valgrind should catch the same errors so it ought to work just as well but maybe not.  Valgrind is also probably slower because it operates on the binary instead of instrumenting the code during compilation.  Valgrind's benefit is that you don't have to recompile and it also works on files that aren't in c.

We should use both as appropriate.

ooxi:
I asked @CiscoTalos if they would be willing to check the poc. Would be great if we could include it. Otherwise I'll try to reproduce myself.

Do you happen to have experience with a fuzzer like asl? My gut feeling is that we have dozens of problems in the parsing code

eyal0:
I know what a fuzzer is.  That's about the extent of my knowledge!  ;-)

Maybe the Cisco people would like to submit a PR?  I'd be happy to help.

ooxi:
I decided to merge this fix in order to defend against CVE-2021-40394, even if the fix is not yet verified by a CI test.

CiscoTalos:
Would you like our Researcher to test this patch before the issue is publicly disclosed?

ooxi:
If you could verify this patch, it would be of tremendous help to us @CiscoTalos!

CiscoTalos:
The patch you proposed fixes the issue described in TALOS-2021-1405.  
Please also take a look at a new issue (#80) which is in this same function.

ooxi:
> The patch you proposed fixes the issue described in TALOS-2021-1405.

thank you very much for checking the patch


> Please also take a look at a new issue (#80) which is in this same function.

We will have a look at this and propose a fix soon

CiscoTalos:
Is there a proposed fix/release date for this issue ?

CiscoTalos:
Is there a fix for #80 ?

ooxi:
Unfortunately, we do not yet have a fix for #80. Please continue with public disclosure

PR #62, opened 2021-11-04
Since nearly 20 years issues are tracked using an issue tracker

ooxi:
It almost hurts deleting such an ancient file :-( but it currently does not serve a purpose in the root directory.


eyal0:
gerbv, welcome to the 21st century!  🤣

PR #69, opened 2021-11-13
Various minor fixups

tijuca:
Mostly all these commits have been created within the Debian packaging since I had some hands on the Debian packaging of gerbv.

The root for the commits were the Lintian QS tool which is used within Debian to detect various types of issues within the packaging.

(I've also a update for the autotool configuration files in the pipe. But this requires at least commit 39351b22a6a1a4aa3409e414e883c32043508f41 applied.)


ooxi (commented):
Is this Debian specific or is UTF-8 the default for all `.desktop` files?

tijuca:
Hi,

UTF-8 is mandatory within Debian as default.
But also within the specification for desktop files.
https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#basic-format

So defining the Encoding as UTF-8 is useless (as this **is** the default), setting something different encoding might be helpful on Windows Systems for example. But even there UTF-8 is the default newer days.

ooxi:
Thank you very much for providing these fixes! In the future I would prefer separate pull requests, but this is fine since it's a collection of the changes you have done over time :-)

PR #70, opened 2021-11-16
Live update of viewer via x&y translation spinbuttons on layer edit window

Rix:


ooxi:
Such a simple, yet brilliant QOL improvement! Would love to see this done for scale and rotation, too :+1:

Issue #56, opened 2021-11-03
Gerbv RS-274X format aperture macro variables out-of-bounds write vulnerability (TALOS-2021-1404)


CiscoTalos:
TALOS-2021-1404   
CVE-2021-40393   

Gerbv RS-274X format aperture macro variables out-of-bounds write vulnerability

### Summary

An out-of-bounds write vulnerability exists in the RS-274X aperture macro variables handling functionality of Gerbv 2.7.0 and dev (commit b5f1eacd). A specially-crafted gerber file can lead to code execution. An attacker can provide a malicious file to trigger this vulnerability.

### Tested Versions

Gerbv 2.7.0   
Gerbv dev  (commit b5f1eacd)   
Gerbv forked dev (commit 71493260)   

### Product URLs

[https://sourceforge.net/projects/gerbv/](https://sourceforge.net/projects/gerbv/)

### CVSSv3 Score

10.0 - CVSS:3.0/AV:N/AC:L/PR:N/UI:N/S:C/C:H/I:H/A:H

### CWE

CWE-119 - Improper Restriction of Operations within the Bounds of a Memory Buffer

### Details

Gerbv is an open-source software that allows to view RS-274X Gerber files, Excellon drill files and pick-n-place files. These file formats are used in industry to describe the layers of a printed circuit board and are a core part of the manufacturing process.

Some PCB (printed circuit board) manufacturers use software like Gerbv in their web interfaces as a tool to convert Gerber (or other supported) files into images. Users can upload gerber files to the manufacturer website, which are converted to an image to be displayed in the browser, so that users can verify that what has been uploaded matches their expectations. Gerbv can do such conversions using the `-x` switch (export). For this reason, we consider this software as reachable via network without user interaction or privilege requirements.

Gerbv uses the function `gerbv_open_image` to open files. In this advisory we're interested in the RS-274X file-type.

    int
    gerbv_open_image(gerbv_project_t *gerbvProject, char *filename, int idx, int reload,
                    gerbv_HID_Attribute *fattr, int n_fattr, gboolean forceLoadFile)
    {
        ...        
        dprintf("In open_image, about to try opening filename = %s\n", filename);
        
        fd = gerb_fopen(filename);
        if (fd == NULL) {
            GERB_COMPILE_ERROR(_("Trying to open \"%s\": %s"),
                            filename, strerror(errno));
            return -1;
        }
        ...
        if (gerber_is_rs274x_p(fd, &foundBinary)) {                                 // [1]
[...]

PR #73, opened 2021-11-19
Fix CVE-2021-40393

ooxi:
A possible out-of-bounds write vulnerability was discovered by the @CiscoTalos team in issue #56. This patch mitigates the issue by adding bounds checking.

### DIFF
--- a/src/gerber.c
+++ b/src/gerber.c
@@ -28,6 +28,7 @@
 
 #include "gerbv.h"
 
+#include <stddef.h>
 #include <stdlib.h>
 #include <string.h>
 #include <limits.h>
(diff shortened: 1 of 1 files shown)


ooxi:
@eyal0 could you do me a solid and take a look at my out-of-bounds checking logic? There might be a signed integer underflow UB since `ip->data.ival - 1` is not checked.

ooxi:
@CiscoTalos could you please take a look whether this patch seems reasonable? I'm unsure whether `ip->data.ival - 1` is still an attack vector since `ip->data.ival - 1` might underflow

CiscoTalos:
`ip->data.ival - 1` will underflow if `ip->data.ival` is 0, but the `if` condition that happens next will terminate the program. So the suggested patch fixes the issue.

ooxi:
Thank you very much for the verification. The fix is now included in main branch and will be released with the next gerbv version (probably in the upcoming week).

CiscoTalos:
Thank you. We prefer 1-2 business days notice of the release date so we can coordinate on our end.

ooxi:
@CiscoTalos I aim to create a release this weekend. Does this suit you?

PR #101, opened 2022-03-02
Altium number format

### LINKED ISSUE #97: gerbv doesn't parse Altium Excellon number format comments
Here is an Excellon file that confuses gerbv's format auto-detection: 
[LimeSDR-QPCIe_1v2-RoundHoles.TXT](https://github.com/gerbv/gerbv/files/8009963/LimeSDR-QPCIe_1v2-RoundHoles.TXT)

This file is generated by Altium. Altium puts the number format of an excellon file inside a comment in the header like this:
```
;FILE_FORMAT=4:4
```

I think it would be worth it for gerbv to parse these comments and use them before guessing the number format.

eyal0:
This fixes #97


ooxi:
Thank you very much! We should do a new release including this feature

eyal0:
Looking at https://coveralls.io/builds/47762774/source?filename=src%2Fdrill.c#L1549 I see that I have some uncovered lines.  So I have pushed a new version with a modified test that will get better coverage.

eyal0:
We see that coverage is improved with the new version: https://coveralls.io/builds/47790302/source?filename=src%2Fdrill.c#L1549

Issue #115, opened 2022-05-21
Improve information for developers


ooxi:
[Information for developers](https://github.com/gerbv/gerbv#information-for-developers) should link to the API documentation as well as the example programs.

ooxi:
Fixed by @jesusdoza in pull request #120

PR #120, opened 2022-05-21
Improve information for developers #115

jesusdoza:
added links to API and example programs


ooxi:
Thank you! What do you think about folding the list of examples using [GitHub Markdown Spoilers](https://gist.github.com/jbsulli/03df3cdce94ee97937ebda0ffef28287)? That way the length of the README does not grow too much :-)

jesusdoza:
ok il change the links to the new URL and add markdown spoilers

PR #125, opened 2022-06-03
Improve const correctness

ooxi:
When using libgerbv from C++ usability can be improved by marking effective const variables as `const`


ooxi:
Thanks for the feedback @eyal0  and @spe-ciellt :-) will release the next RC with the CVE fix soon

Issue #80, opened 2021-11-22
Security advisory TALOS-2021-1416


CiscoTalos:
TALOS-2021-1416   
CVE-2021-40402   

Gerbv RS-274X aperture macro multiple outline primitives out-of-bounds read vulnerability

### Summary

An out-of-bounds read vulnerability exists in the RS-274X aperture macro multiple outline primitives functionality of Gerbv 2.7.0 and dev (commit b5f1eacd), and Gerbv forked 2.7.1 and 2.8.0. A specially-crafted gerber file can lead to information disclosure. An attacker can provide a malicious file to trigger this vulnerability.

### Tested Versions

Gerbv 2.7.0   
Gerbv forked 2.7.1   
Gerbv forked 2.8.0   
Gerbv dev  (commit b5f1eacd)   

### Product URLs

[https://sourceforge.net/projects/gerbv/](https://sourceforge.net/projects/gerbv/)

### CVSSv3 Score

9.3 - CVSS:3.0/AV:N/AC:L/PR:N/UI:N/S:C/C:L/I:N/A:H

### CWE

CWE-755 - Improper Handling of Exceptional Conditions

### Details

Gerbv is an open-source software that allows users to view RS-274X Gerber files, Excellon drill files and pick-n-place files. These file formats are used in industry to describe the layers of a printed circuit board and are a core part of the manufacturing process.

Some PCB (printed circuit board) manufacturers use software like Gerbv in their web interfaces as a tool to convert Gerber (or other supported) files into images. Users can upload gerber files to the manufacturer website, which are converted to an image to be displayed in the browser, so that users can verify that what has been uploaded matches their expectations. Gerbv can do such conversions using the `-x` switch (export). For this reason, we consider this software as reachable via network without user interaction or privilege requirements.

Gerbv uses the function `gerbv_open_image` to open files. In this advisory we're interested in the RS-274X file-type.

    int
    gerbv_open_image(gerbv_project_t *gerbvProject, char *filename, int idx, int reload,
                    gerbv_HID_Attribute *fattr, int n_fattr, gboolean forceLoadFile)
    {
        ...        
        dprintf("In open_image, about to try opening filename = %s\n", filename);
        
        fd = gerb_fopen(filename);
        if (fd == NULL) {
            GERB_COMPILE_ERROR(_("Trying to open \"%s\": %s"),
                            filename, strerror(errno));
            return -1;
        }
        ...
[...]

PR #126, opened 2022-06-03
RFC: How to patch CVE-2021-40402

ooxi:
Issue #80 has identified multiple out of bound accesses regarding [gerbv_simplified_amacro_t::parameter](https://github.com/gerbv/gerbv/blob/main/src/gerbv.h#L422).

I have banged my head against this problem for a couple of days, but could not find a good way without playing wack a mole. At least `gerbv_simplified_amacro_t::parameter` and `gerbv_aperture_t::parameter` are vulnerable and accessed throughout the program.

Three ways of accessing these arrays have been identified:

* `val = sam->parameter[idx]`: No problem here, access could be replaced by a simple inline function doing bounds checking
* `sam->parameter[idx] += val`: Cannot be replaced with an inline function, unless returning a pointer and doing `*sam_parameter(sam, idx) += val`. Decided against returning a pointer and instead using a macro
* `*double val = sam_parameter + idx`: Could be replaced with an inline function returning a pointer, as well

@eyal0 / @spe-ciellt: Do you have input on how to patch this vulnerability? I would prefer replacing all access sites at once instead of trying manual index validation where necessary.

### DIFF
--- a/src/gerb_safety.h
+++ b/src/gerb_safety.h
@@ -1,0 +1,28 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+#ifndef GERBV_GERBV_SAFETY
+#define GERBV_GERBV_SAFETY
+
+#include <assert.h>
+#include <stddef.h>
+
+#include "gerbv.h"
+
+inline size_t gerbv_simplified_amacro_parameter_idx(gerbv_simplified_amacro_t const* amacro, size_t idx) {
+    static_assert(
+        (sizeof(amacro->parameter) / sizeof(amacro->parameter[0])) == APERTURE_PARAMETERS_MAX,
+        "Statically known gerbv_simplified_amacro parameter array size has changed");
+
+    if (idx < APERTURE_PARAMETERS_MAX) {
+        return idx;
+    }
+
+    GERB_FATAL_ERROR(/*_*/("invalid gerbv_simplified_amacro parameter index"));
+    return 0;
+}
+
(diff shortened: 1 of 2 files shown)


eyal0:
It's made extra hard because we can't use c++.  With c++, you could just override the definition of the operator and maybe put a bounds check there.

This is a complex one.  I'd be tempted to rename `parameter` to something else just so that the compiler can ensure that we didn't miss one.  But I don't otherwise have any brilliant ideas.

spe-ciellt:
I was thinking exactly as @eyal0 . I have nothing against C++, actually I would prefer it. But then the first step would be to make gerbv at least compile using g++ (or clang), and _then_ fix the issue at hand. But that feels like a major task.

Either way, sounds like you can't really get away without patching every place. One solution would be to "hide" the functionality in a C++ class way. Then it wouldn't compile until you fixed all the places, which is good in a sense that then the compiler would find all places to update. Just realized this is about the same as @eyal0 said. So no, no brilliant ideas here either.

ooxi:
Thanks for your feedback!

Using C++ for certain data structures would simplify many aspects. However, this would also break API for libgerbv and certainly transform gerbv as a whole. I'm not ruling out moving to C++ in the long run, but for this it seems like too big of a change.

Most likely I'll proceed by temporary renaming `parameter` in order to make the patch more readable. I do not like renaming it altogether since it's part of the exposed API.

PR #132, opened 2022-09-21
Update BUGS file to point to issue tracker

### LINKED ISSUE #129: Is the BUGS file current?
Hello!

I found the file `BUGS` in the root directory of the repository. It apparently lists a number of bugs. Are the bugs in the list still valid?
They are also built into the program when it is built. If they are valid, should they be built into the program? Seems more  like they should be here and then point to here in the "About" in the program.

ooxi:
The BUGS file was no longer kept up to date since half a decade ago. Therefore it's more user friendly to point to the issue tracker.

I have not updated the translation files. Will fix #129

### DIFF
--- a/BUGS
+++ b/BUGS
@@ -1,11 +1,2 @@
-Currently KNOWN BUGS
-
-All OS:
-
-If layer orientation modified with rotation and rendering mode is 'Fast' or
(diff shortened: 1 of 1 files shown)


eyal0:
Good idea.  No need to have two lists of bugs!

PR #133, opened 2022-09-21
fix: Free the memory allocated in aperture macro comments.

eyal0:
The return of `gerb_fgetstring` must be freed even though it is not used.  `gerb_fgetstring` may return NULL but `free(NULL)` is allowed and does nothing.

A comment is added into an aperture macro of one of the test files. That file is added to the list of examples to be run with valgrind.


ooxi:
Thanks!

PR #135, opened 2022-10-26
Fix rotation

### LINKED ISSUE #134: Turning the Gerber Fail
If you rotate the file 90 degrees, you cannot save the result. When you reopen, the file returns to its original position. When you change the coordinates or rotate the file, the settings window cannot be moved and it closes the original picture

eyal0:
This fixes #134


ooxi:
This is amazing, thank you!

PR #136, opened 2022-10-27
ci: Improve run_tests.sh --regen

eyal0:
This make it a little easier to use the regen feature of `run_tests.sh`.


ooxi:
Thank you!

PR #137, opened 2022-10-27
ci: Upload gerbv executables as artifacts

eyal0:
Users will be able to download the executable directly from the CI actions.  This is helpful for providing binaries for users to test, for example.


ooxi:
Thank you!

Issue #26, opened 2021-09-23
Unify header include guards


ooxi:
libgerbv currently uses multiple different styles of include guards:

* Invalid: Include guards like `__COMMON_H__` which collide with reserved identifiers (7.1.3)
* Name-Clash: Short include guard defines like `CSV_H` which might collide with other projects, since they don't use a project prefix
* Missing: No include guard present, which might lead to redefinitions
* Prefixed: Include guards like `GERBV_LAYERTYPE_PICKANDPLACE_H` which are both valid identifiers and do not risk collission with other projects
* Vendored: Header copied from or generated by third party

I propose to add prefixed include guards (like `GERBV_CSV_H`) to all headers unless they are vendored

## Pros

* No risk of include guard collissions due to project specific prefix
* No risk of redefinitions due to missing include guards
* No risk of unexpected behaviour since we are not using reserved identifiers

## Cons

* We change the defined symbols, which might break source compatibility for very fragile users

## Alternatives

We could use `#pragma once` instead, which is supported by all platforms we care about. Nevertheless I would refrain from using a non standard feature.

## Overview

| Header                                     | Include guard                    | Category   |
| ------------------------------------------ | -------------------------------- | ---------- |
| [`amacro.h`](https://github.com/gerbv/gerbv/blob/main/src/amacro.h)                 | `AMACRO_H`                       | Name-Clash |
| [`attribute.h`](https://github.com/gerbv/gerbv/blob/main/src/attribute.h)           | `ATTRIBUTE_H`                    | Name-Clash |
| [`callbacks.h`](https://github.com/gerbv/gerbv/blob/main/src/callbacks.h)           | Missing                          | Missing    |
| [`common.h`](https://github.com/gerbv/gerbv/blob/main/src/common.h)                 | `__COMMON_H__`                   | Invalid    |
| [`csv_defines.h`](https://github.com/gerbv/gerbv/blob/main/src/csv_defines.h)       | `DEFINES_H`                      | Name-Clash |
| [`csv.h`](https://github.com/gerbv/gerbv/blob/main/src/csv.h)                       | `CSV_H`                          | Name-Clash |
| [`draw-gdk.h`](https://github.com/gerbv/gerbv/blob/main/src/draw-gdk.h)             | `DRAW_GDK_H`                     | Name-Clash |
| [`draw.h`](https://github.com/gerbv/gerbv/blob/main/src/draw.h)                     | `DRAW_H`                         | Name-Clash |
[...]

PR #142, opened 2022-10-30
Unify include guards

ooxi:
All non vendored headers now use unified prefixed include guards.

See [issue #26](https://github.com/gerbv/gerbv/issues/26) for motivation and discussion.


ooxi:
Will have to check the failed build in ci-coverage

PR #145, opened 2022-11-02
Debug build

eyal0:


spe-ciellt:
Debug builds are now done via CMake flags

PR #153, opened 2022-11-18
Checkout specific lcov release to avoid random build failure

ooxi:
This way we can decide when to update the dependency. Fixes issue #152


eyal0:
Smart!

PR #156, opened 2022-11-28
Turn off autodetect after FILE_FORMAT

### LINKED ISSUE #155: Numeric format incorrect when reading Altium 14 NC Drill file
gerbv is misreading the numeric format in an Altium Excellon drill file.  The top of an example is here:

```
M48
;Layer_Color=9474304
;FILE_FORMAT=2:3
INCH,TZ
;TYPE=PLATED
T1F00S00C0.028
T2F00S00C0.035
T3F00S00C0.040
T4F00S00C0.047
T5F00S00C0.051
T6F00S00C0.072
T7F00S00C0.125
%
T01
X1875Y350
X1965Y460
etc.
```

Tracing execution in drill.c, it is correctly parsing the ```FILE_FORMAT=2:3``` line, but the subsequent ```INCH,TZ``` line resets to trailing zero suppression with 4 decimals.  The file then reads at 1/10 of the proper scale.

I don't know if Altium is at fault, but without a detailed knowledge of the standard, it looks incorrect to set TZ because the rest of file is obviously not suppressing trailing zeros (a dumb concept, but there you have it).  But an Excellon guru needs to look at it.

If, in gerbv viewer, the format is manually reset to 3 decimals (no autodetect), it reads at correct scale.  

Possibly, ```state->autod``` should be reset if the ```FILE_FORMAT``` line is read, or another flag should be set to prevent any change to the format.

This may be related to issue #94.

eyal0:
This fixes #155


ooxi:
Great PR, especially first adding a test case and then fixing it

eyal0:
> Great PR, especially first adding a test case and then fixing it

That was no accident! It allows us to see the effect of the change together with the new code. I like it that way, add a test and then add code.

ooxi:
> That was no accident! It allows us to see the effect of the change together with the new code.

Yes I appreciate the extra effort it takes to do it this way. Makes the commits much easier to reason about :-)

Issue #157, opened 2022-11-28
Enhancement request: generalize drawing.


meantaipan:
Currently, the code basically relies on Cairo for rendering (pixmap or vector format).  I would like to be able to generate output suitable for use with OpenSceneGraph and similar 3D rendering libraries.  This would generate board views similar to what most EDA tools can do, that can be flipped around with the mouse with realistic color, thickness etc.

I am prepared to do this work, but I am asking here whether this is the best place/repository to do it.  Thoughts?

The approach I have in mind at present is to change direct calls to Cairo (e.g. ```cairo_move_to()```) to indirect calls via function pointers: ```draw_impl->move_to()```.  Instead of passing ```cairo_t *cairoTarget``` around, pass a struct that is one additional level of abstraction away from Cairo.  Naturally, a default object would be set up to use Cairo, so that library functionality is unchanged by default, however a library user could create their own draw implementation which uses OSG instead of Cairo, provide they stick to Cairo-compatible semantics.

Another enhancement is to support IPC-D-356A test data (I already use this to run a soldering robot but would be nice to integrate graphically), and this in turn suggests support for RS274-X2 (attributes) as a way to store the data.

Open to suggestions...

eyal0:
`pcb2gcode` uses gerbv as just a parser of the file and converts the entirety of a gerber file into a ["well known text"](https://en.wikipedia.org/wiki/Well-known_text_representation_of_geometry), which is a common way of representing lines and shapes and used a lot for GIS stuff (like maps).  WKT is easily converted to SVG files, they are very similar in structure.

The entrypoint for the conversion is here:

https://github.com/pcb2gcode/pcb2gcode/blob/master/gerberimporter.cpp#L735-L738

The `render` function needs to know if you want to fill closed lines (this is useful for things that are drawn as an outline, such as edge cuts layers).  You can choose to either return just shapes or shapes and lines, where the lines are for traces that were cut out with a circular tool.

The tool will use libgeos if it is installed and if not, it will use boost.  Either way, boost must be installed.

The input is the gerber file opened by libgerbv.

If you check out the repo, run `autoconf -fvi && ./configure && make -j 20 check`.  Then I suggest that you look at the `gerberimporter_tests` because it is testing *just* the gerberimporter.

Try adding this line to the top of the file `gerberimporter_tests.cpp`:

```c++
#include <boost/geometry/io/wkt/write.hpp>
```

And then add this on line 205, right after the render:

```c++
[...]

PR #160, opened 2022-12-15
IPC-D-356A testpoint data support plus RS274-X2 attributes.

meantaipan:
This is for issue #157.

Hopefully you can squash the commit history if you pull.  Otherwise, let me know and I'll make a squashed branch.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -2,6 +2,9 @@
 Release Notes for gerbv-2.9.6-dev
 ========================================================================
 
+-gerbv:     Turn off autodetect after FILE_FORMAT (PR#156 by @eyal0 fixing issue
+            #155 reported by @meantaipan)
+
 
 ========================================================================
 Release Notes for gerbv-2.9.5


ooxi:
The ci-coverage failing is not an issue, but somehow the tests itself do not seem to run. Did they pass in your fork?

meantaipan:
@ooxi yes all tests passed, but I had to regen some of the "golden" PNG files.

This test failed on the pick and place example.  This was one I had to regen.  It is probably failing for you because of minor differences when rendering text (exact typeface used, pixel-level differences etc.)

This is a weakness in the automated tests that probably needs to be resolved.  Rather than comparing the test output with a single golden PNG, it needs to compare with a set of them, and only fail if none match.  So if the test name is 'foo' then it will compare with foo.png, foo$ubuntu16.04.png, foo$raspberrypi.png and so on.

For now, I would bypass the pick and place test.  I think it's the only one that writes text.

eyal0:
I'm not able to build your code on my computer yet.  I'll keep trying...

eyal0:
@meantaipan Can you please try the following commands:

```bash
git co meantaipan/main
git rebase gerbv/main -x "make -j 20 clean && make -j 20 && (make check || (cd test ; ./run_tests.sh --regen ; cd .. ; make check))"
```

This will check that every commit is code that can be compiled and it will also update the golden files at each step.  This will show us which change is the one that causes the golden files to change.

The golden files work on my computer and also in CI and the CI platform hasn't changed so I would expect that the tests would continue to pass, right?

meantaipan:
@eyal0 Only change to rendered output is for the pick and place example, which writes text.  Everything not involving text should be exactly the same.

eyal0:
@meantaipan After the rebase stuff completes then I'll push the update and we'll see which change made the difference, yes?

meantaipan:
@eyal0 OK, I'm curious to know.  I'm certainly not a git guru.  My version doesn't even understand "git co".

eyal0:
> @eyal0 OK, I'm curious to know. I'm certainly not a git guru. My version doesn't even understand "git co".

Oops, haha, `co` is short for `checkout`!

I see that the tests are failing on valgrind, too, which means that there is a new memory leak.  We'll look for that also.

meantaipan:
@eyal0 That's embarrassing.  Don't worry, I know how to use valgrind, so I can look into that (but not until this evening California time, about 8 hours away).

eyal0:
Oops I did a bad push.  It's confusing me very much that your branch and this branch are both called main!

Let's not call the feature branch `main` please

meantaipan:
@eyal0 Yeah my bad.  My local repo uses a different branch name, but for some dumb reason I pushed to main on my github fork.  Know better now.

eyal0:
I re-opened on a different branch because my bad push closed this one.  I also see some warnings.  Anyway, take a look at #161 and address those.  And then you can rebase and put it on your own branch in your own fork of the repo if you want, no problem.

PR #161, opened 2022-12-16
IPC-D-356A testpoint data support plus RS274-X2 attributes.

eyal0:
From #160 which I accidentally closed and cannot re-open, oops.  @meantaipan


eyal0:
I run this command to see if each individual commit is passing tests:

```bash
git rebase main -x "./autogen.sh && ./configure --prefix=/home/esoha/.local && make -j 20 clean && make -j 20 && (make check || (cd test ; ./run_tests.sh --regen ; cd .. ; make check))"
```

eyal0:
The first failure is in 0bc88c7179c04e79e4b97aef909d298e96830f76 found by valgrind.

```
==2806446== 1 bytes in 1 blocks are definitely lost in loss record 2 of 279
==2806446==    at 0x483979B: malloc (vg_replace_malloc.c:393)
==2806446==    by 0x507AD48: g_malloc (in /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0.6600.8)
==2806446==    by 0x5094B09: g_strndup (in /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0.6600.8)
==2806446==    by 0x486D90E: gerbv_set_render_options_for_file (gerbv.c:1273)
==2806446==    by 0x486DB85: gerbv_render_all_layers_to_cairo_target (gerbv.c:1305)
==2806446==    by 0x486049A: gerbv_export_png_file_from_project (export-image.c:87)
==2806446==    by 0x11977B: main (main.c:1067)
```

src/gerbv.c
```diff
@@ -968,12 +1227,69 @@ gerbv_render_all_layers_to_cairo_target_for_vector_output (
...
+	else if (pts < tmin)
+	        renderInfo->textSizeInch = 0.;
+
+        //printf("SROF: text=%g,%g,%g;  pts=%g, scaleFactorX=%g, textSizeInch=%g\n", 
+        //        tmin, tmax, tmils, pts, renderInfo->scaleFactorX, renderInfo->textSizeInch);
+
+        // default to file color
+        renderInfo->textColor = fileInfo->color;
+        text_color = x2attr_get_field_or_last(0, 
+                x2attr_get_project_attr_or_default(gerbvProject, "text-color", ""));
```
eyal0, line 1302: The memory leak is around here.  Maybe `text_color` needs to be freed?

src/x2attr.c
```diff
@@ -773,3 +794,271 @@ x2attr_set_project_attr(gerbv_project_t * project,
...
+    // If the array has zero size, an empty string is returned.
+    unsigned l, i;
+    char * s, * p;
+    g_assert(utf8_array);
+    
+    if (!utf8_array->len)
+        return g_strdup("");
+    
+    for (i = 0; i < utf8_array->len; ++i)
+        l += _utf8_to_file(((char **)utf8_array->data)[i], NULL);
```
eyal0, line 964:
```
x2attr.c: In function ‘x2attr_utf8_array_to_file’:
x2attr.c:965:11: error: ‘l’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
  965 |         l += _utf8_to_file(((char **)utf8_array->data)[i], NULL);
      |           ^~
```

src/gerbv.c
```diff
@@ -531,38 +536,278 @@ gerbv_open_image(gerbv_project_t *gerbvProject, gchar const* filename, int idx,
...
+            )
+            ++instance;
+    }
+    
+    // If target layertype is RS274-X, then parse any relevant project options that assign X2 attributes
+    // etc.  The "layers" arg attribute is particularly useful for assigning a function to each
+    // layer.  It also indicates how many layers to expect.  E.g. if --layers=1,2,3,4 then we know it's a 4-layer
+    // board and the bottom layer is #4.
+    
+    if (layertype == GERBV_LAYERTYPE_RS274X) {
```
eyal0, line 590:
```
gerbv.c: In function ‘gerbv_open_image’:
gerbv.c:587:8: error: ‘layertype’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
  587 |     if (layertype == GERBV_LAYERTYPE_RS274X) {
      |        ^
```

meantaipan:
@eyal0 The memory leak is probably not real.  For attribute keys and data, I am relying on GLib interning the strings, hence everything is a const char * without needing to free.  To quieten valgrind, I will need to see if GLib has an API to tell it to free all the interned strings.  We'll have to call that at the end of main().  From a pure library perspective, we might need to define a "cleanup" function, although the lifetime of the interned strings must remain for the entire application, since we can't pick and choose which ones to clean up.

The uninit variables are a fair catch.  Easy to fix.

meantaipan:
BTW, the source line numbers don't seem to line up in the valgrind output.  And there's not even any instance of calling g_strndup() from gerbv_set_render_options_for_file().  Not directly, anyway.

meantaipan:
Argh!  There's no default layertype enum to set!  Have to use a cast.

meantaipan:
Looks like GLib interning uses reference counting.  So technically I would have to unref each string instead of leaving it to a global cleanup function.  But maybe there's a deeper problem with that one orphaned byte.  I'll run valgrind myself to see if I can't clean it up a bit more.

eyal0:
the memory leak and errors seem real.  Here's a solution:

```diff
diff --git a/src/gerbv.c b/src/gerbv.c
index 3d0b21a..37d4793 100644
--- a/src/gerbv.c
+++ b/src/gerbv.c
@@ -562,6 +562,10 @@ gerbv_open_image(gerbv_project_t *gerbvProject, gchar const* filename, int idx,
     else if (gerber_is_rs274d_p(fd)) {
         ftype = FILE_TYPE_RS274D;
         layertype = GERBV_LAYERTYPE_RS274X;
+    } else {
+        GERB_COMPILE_ERROR(_("Unrecognized layertype in file \"%s\": %s"),
+                           filename, strerror(errno));
+        return -1;
     }
         
     
@@ -1231,10 +1235,10 @@ gerbv_render_all_layers_to_cairo_target_for_vector_output (
 void
 gerbv_set_render_options_for_file (gerbv_project_t *gerbvProject, gerbv_fileinfo_t *fileInfo, gerbv_render_info_t *renderInfo)
 {
-        const char * text_min;
-        const char * text_max;
-        const char * text_mils;
-        const char * text_color;
+        char * text_min;
+        char * text_max;
+        char * text_mils;
+        char * text_color;
         gfloat tmin, tmax, tmils, pts;
 
         text_min = x2attr_get_field_or_last(0, 
@@ -1250,6 +1254,9 @@ gerbv_set_render_options_for_file (gerbv_project_t *gerbvProject, gerbv_fileinfo
         sscanf(text_min, "%g", &tmin);
         sscanf(text_max, "%g", &tmax);
         sscanf(text_mils, "%g", &tmils);
+        g_free(text_min);
+        g_free(text_max);
+        g_free(text_mils);
         
 	// Set the target text size (em square) in inches.
 	renderInfo->textSizeInch = 0.001 * tmils;
@@ -1278,7 +1285,7 @@ gerbv_set_render_options_for_file (gerbv_project_t *gerbvProject, gerbv_fileinfo
     	                          text_color,
     	                          FALSE,
     	                          "text");
-	
+        g_free(text_color);
 }
```

We need to `g_free` the result of `x2attr_get_field_or_last`, which is also explained in the comment for that function.

src/gerbv.c
```diff
@@ -623,6 +868,17 @@ gerbv_open_image(gerbv_project_t *gerbvProject, gchar const* filename, int idx,
 	parsed_image = NULL;
     }
     
+    
+    if (parsed_image) {
+        // Assign image (file) attributes if applicable
+        if (attrbuf[0])
```
eyal0, line 878: ==2924607== Conditional jump or move depends on uninitialised value(s)
==2924607==    at 0x486B96C: gerbv_open_image (gerbv.c:878)
==2924607==    by 0x486C477: gerbv_open_layer_from_filename_with_color (gerbv.c:253)
==2924607==    by 0x11903D: main (main.c:948)

eyal0: Solved by doing this:

```diff
diff --git a/src/gerbv.c b/src/gerbv.c
index 37d4793..ab6b7b5 100644
--- a/src/gerbv.c
+++ b/src/gerbv.c
@@ -588,6 +588,7 @@ gerbv_open_image(gerbv_project_t *gerbvProject, gchar const* filename, int idx,
     // layer.  It also indicates how many layers to expect.  E.g. if --layers=1,2,3,4 then we know it's a 4-layer
     // board and the bottom layer is #4.
     
+    attrbuf[0] = 0;
     if (layertype == GERBV_LAYERTYPE_RS274X) {
         /* Get *all* layers defined, so can map 'b' to max (bottom) layer.  't' is always 1.
            Copper layer number is always first char: digit(2) 1..9 or t or b.  (Others are not copper).
@@ -685,7 +686,6 @@ gerbv_open_image(gerbv_project_t *gerbvProject, gchar const* filename, int idx,
         g_free(ff);
         // Now that we know what 'b' layer means, go ahead and parse this instance.
         ff = x2attr_get_field_or_default(instance, file_functions, "");
-        attrbuf[0] = 0;
         ffp = ff;
         switch (tolower(*ffp++)) {
         case '1': case '2': case '3': case '4': case '5': case '6': case '7': case '8': case '9': 
```

meantaipan:
There's a few other instances where I forgot to g_free.  grep for all `x2attr_get_field`.  Sheesh.

eyal0:
This is the fix for all the valgrind and compiler warnings so far:

```diff
diff --git a/src/gerbv.c b/src/gerbv.c
index bd5028f..24e5e0f 100644
--- a/src/gerbv.c
+++ b/src/gerbv.c
@@ -565,6 +565,10 @@ gerbv_open_image(gerbv_project_t *gerbvProject, gchar const* filename, int idx,
     else if (gerber_is_rs274d_p(fd)) {
         ftype = FILE_TYPE_RS274D;
         layertype = GERBV_LAYERTYPE_RS274X;
+    } else {
+        GERB_COMPILE_ERROR(_("Unrecognized layertype in file \"%s\": %s"),
+                           filename, strerror(errno));
+        return -1;
     }
         
     
@@ -587,6 +591,7 @@ gerbv_open_image(gerbv_project_t *gerbvProject, gchar const* filename, int idx,
     // layer.  It also indicates how many layers to expect.  E.g. if --layers=1,2,3,4 then we know it's a 4-layer
     // board and the bottom layer is #4.
     
+    attrbuf[0] = 0;
     if (layertype == GERBV_LAYERTYPE_RS274X) {
         /* Get *all* layers defined, so can map 'b' to max (bottom) layer.  't' is always 1.
            Copper layer number is always first char: digit(2) 1..9 or t or b.  (Others are not copper).
@@ -684,7 +689,6 @@ gerbv_open_image(gerbv_project_t *gerbvProject, gchar const* filename, int idx,
         g_free(ff);
         // Now that we know what 'b' layer means, go ahead and parse this instance.
         ff = x2attr_get_field_or_default(instance, file_functions, "");
-        attrbuf[0] = 0;
         ffp = ff;
         switch (tolower(*ffp++)) {
         case '1': case '2': case '3': case '4': case '5': case '6': case '7': case '8': case '9': 
@@ -1256,10 +1260,10 @@ gerbv_set_render_options_for_file (gerbv_project_t *gerbvProject,
                                    gerbv_fileinfo_t *fileInfo, 
                                    gerbv_render_info_t *renderInfo)
 {
-        const char * text_min;
-        const char * text_max;
-        const char * text_mils;
-        const char * text_color;
+        char * text_min;
+        char * text_max;
+        char * text_mils;
+        char * text_color;
         gfloat tmin, tmax, tmils, pts;
 
         text_min = x2attr_get_field_or_last(0, 
@@ -1275,6 +1279,9 @@ gerbv_set_render_options_for_file (gerbv_project_t *gerbvProject,
         sscanf(text_min, "%g", &tmin);
         sscanf(text_max, "%g", &tmax);
         sscanf(text_mils, "%g", &tmils);
+        g_free(text_min);
+        g_free(text_max);
+        g_free(text_mils);
         
 	// Set the target text size (em square) in inches.
 	renderInfo->textSizeInch = 0.001 * tmils;
@@ -1306,7 +1313,7 @@ gerbv_set_render_options_for_file (gerbv_project_t *gerbvProject,
     	                          text_color,
     	                          FALSE,
     	                          "text");
-	
+        g_free(text_color);
 }
 
 
diff --git a/src/x2attr.c b/src/x2attr.c
index 3df76fe..c8167e7 100644
--- a/src/x2attr.c
+++ b/src/x2attr.c
@@ -960,6 +960,7 @@ x2attr_utf8_array_to_file(const GArray * utf8_array)
     if (!utf8_array->len)
         return g_strdup("");
     
+    l = 0;
     for (i = 0; i < utf8_array->len; ++i)
         l += _utf8_to_file(((char **)utf8_array->data)[i], NULL);
     l += utf8_array->len;   // for the commas and null term
```

eyal0:
> There's a few other instances where I forgot to g_free. grep for all `x2attr_get_field`. Sheesh.

It's okay, no problem!  You did a big job, it's expected to have some fixes to it!

eyal0:
I also needed to set `thislayer` to 0 at the start of `gerbv_open_image` to solve another valgrind error about an uninitialized variable.

eyal0:
Tests pass now I think.

meantaipan:
Great! - but what was the issue with `thislayer`?  I can't see how it was used without being init in the default case at line 707.

meantaipan:
@eyal0 Is there somewhere I can pull the patches you made?  I would like to continue to create more test cases for the stuff I was working on.  It would be nice to have the most solid base.

eyal0:
Sure, you could build off this PR. Or compare yours to this. This one is like yours but with small changes to make it pass tests.

meantaipan:
@eyal0 Great!  So, from my Linux box, what's the magic git command to merge this PR?

meantaipan:
Nevermind, I created a remote to your branch.  For some reason it came up with all conflicts rather than a fast forward.

Anyway, trying to run valgrind tests, but the -s option to valgrind is not understood.  What is that supposed to do?  Can't find it in the valgrind documentation.

meantaipan:
Nevermind again, I just upgraded to the latest valgrind.

meantaipan:
I have been testing export in RS274-X format.  It seems there is a problem with layer states not getting set correctly.  For example, the `test-image-polarity-1.gbx` example:

Input:

```
G04 Test image polarity *
G04 Crosshairs should be cut out of a positive background*
G04 Handcoded by Julian Lamb *
%MOIN*%
%FSLAX23Y23*%
%IPNEG*%
%ADD10C,0.050*%

G04 Draw crosshairs *
X-1000Y0D02*
G54D10*
X1000Y0D01*
X0Y-1000D02*
G54D10*
X0Y1000D01*

M02*
```

This gets written out as:

```
%TFLayerMax,1*%
%TFLayerNum,0*%
G04 This is an RS-274x file exported by *
G04 gerbv version 2.9.6-dev~a11dc9 *
G04 More information is available about gerbv at *
G04 https://gerbv.github.io/ *
G04 --End of header info--*
%MOMM*%
%FSLAX46Y46*%
%IPPOS*%
G04 --Define apertures--*
%ADD10C,1.2700*%
G04 --Start main section--*
D10*
G01X-25400000Y0D02*
G01X25400000Y0D01*
G01X0Y-25400000D02*
G01X0Y25400000D01*
M02*
```

So ignoring the x2 attributes (and metric units) which I probably should not be inserting, it has reverted to positive polarity.

eyal0:
IPNEG!  Oh man.  Here's the doc on IPNEG:

```
IP Sets the ‘Image polarity’ graphics state
parameter
This command has no effect in CAD to CAM
workflows. Sometimes used, and then usually as
%IPPOS*% to confirm the default and then it
then has no effect. As it is not clear how
%IPNEG*% must be handled it is probably a
waste of time to try to fully implement it, and
sufficient to give a warning on a %IPNEG*% and
skip it.
Deprecated in 2013
```

You're not supposed to make images that are just negatives.  gerbv is supporting it, I guess.

eyal0:
That's on page 170, BTW: https://www.ucamco.com/files/downloads/file_en/456/gerber-layer-format-specification-revision-2022-02_en.pdf?7b3ca7f0753aa2d77f5f9afe31b9f826

meantaipan:
Yes, I realized that a lot of features supported by gerbv are deprecated, but obviously we want to be able to parse old files.  In the process of working on issue #162, I have added support for writing out IPNEG (and a whole bunch of other stuff that was not implemented, like step and repeat).  I'm almost complete with that, so don't go and duplicate that effort.  It will be in an upcoming PR.

I think that being able to export a file which behaves identically when imported is a good feature and certainly should help with automatic export validation.  But it's too complicated a task to upgrade deprecated features to do the Right Thing, so for now just echo out the deprecated constructs.

meantaipan:
Since the standard notes that some of these deprecated constructs are hazardous (such as Mirror Image) I'll add a deprecation warning that will be logged if any are encountered in the input.  Might add `--noxxx` options etc. (a-la gcc warning suppression) just for regression testing.  What do you think?

eyal0:
I only did the rebase.  But this thing is so massive that I don't even know where to begin!  I'm able to look at it more next week, though.  Is there a way to break this up into multiple commits so that the review is easier?

ooxi:
@meantaipan I agree with eyal0. It would be great if we could merge the commits in smaller chunks.

For example, is [143ae2bfe16bf03725ea1ebd5bc10ba268afccd1](https://github.com/gerbv/gerbv/pull/161/commits/143ae2bfe16bf03725ea1ebd5bc10ba268afccd1) a self contained feature? If so, I would prefer working on getting that merged

meantaipan:
Yes, I tried to commit at reasonable points, so hopefully it would make your life easier to merge commits that way.  Unfortunately, I did not run the regression tests at each step so no guarantee that you'll be able to test each step that way.

So is the main problem that the change is so big that you are getting numerous conflicts with other changes (which I was not aware of, because I based from the main branch)?  I feel your pain.

I do recall asking up front about how often I should commit my own development steps, and got the impression you all don't want it too granular, so I tried to keep it to fairly complete chunks of work.  With the benefit of hindsight, I now see that I should have made it a little finer grained.  Sorry about that.  Put it down to being a newcomer to open source cooperation.  I'm used to having a fairly linear development cycle, since it's normally just me.

ooxi:
@meantaipan not having a separate commit for every little step is fine, we should just try to merge one feature after the other :-)

Let's split 143ae2bfe16bf03725ea1ebd5bc10ba268afccd1 into a separate pull request and continue from there.

src/x2attr.c
```diff
@@ -0,0 +1,649 @@
...
+  
+  NOTE: const char * parameters must be interned strings!  Call with g_intern_string(foo) if in doubt.
+  
+*******************/
+
+static const char *
+_find_value(const x2attr_state_t * s, const char * key)
+{
+    unsigned i;
+    if (s)
```
eyal0, line 115: I prefer to always use braces.  Apple had a very serious bug when they didn't do this: https://nakedsecurity.sophos.com/2014/02/24/anatomy-of-a-goto-fail-apples-ssl-bug-explained-plus-an-unofficial-patch/ .  But I realize that some of the code here doesn't do it so...

src/x2attr.c
```diff
@@ -0,0 +1,649 @@
...
+
+/******************
+  Local functions on xattr_state
+  
+  NOTE: const char * parameters must be interned strings!  Call with g_intern_string(foo) if in doubt.
+  
+*******************/
+
+static const char *
+_find_value(const x2attr_state_t * s, const char * key)
```
eyal0, line 112: Place this function after `_find_index` below and write:

```c
int index = _find_index(s, key);
return index == -1 ? NULL : s->t[i].value;
```

That's shorter, right?

src/x2attr.c
```diff
@@ -0,0 +1,649 @@
...
+typedef struct x2attr_state {
+    gboolean sorted;        // TRUE when keys are sorted in canonical order.  This is
+                            // only set when _sort() called.  Reset on any modification.
+    unsigned size;          // Number of elements in following
+    struct x2attr_state * next; // Aperture attributes if this is for net (object), etc.
+    struct {
+        const char * key;   // Pointer to interned UTF-8 null terminated string.
+                            // Sort by key if 'sorted'.
+        const char * value; // Pointer to interned UTF-8 null terminated string
+    } t[1]; // 1st of 'size' elements.
```
eyal0, line 100: Why is it called `t`?  Could we be more descriptive with the name?  Like `attributes` or something?

src/x2attr.c
```diff
@@ -0,0 +1,649 @@
...
+x2attr_get_image_attr(const gerbv_image_t * image, const char * key)
+{
+    return _find_value(image->attrs, g_intern_string(key));
+}
+
+const char * 
+x2attr_get_image_attr_or_default(const gerbv_image_t * image, 
+                const char * key, const char * dflt)
+{
+    const char * v = _find_value(image->attrs, g_intern_string(key));
```
eyal0, line 648:
```c
const char * v = x2attr_get_image_attr(image, key);
```

You already defined that function, you might as well use it!

src/x2attr.c
```diff
@@ -0,0 +1,649 @@
...
+x2attr_get_aperture_attr(const gerbv_aperture_t * aperture, const char * key)
+{
+    return _find_value(aperture->attrs, g_intern_string(key));
+}
+
+const char * 
+x2attr_get_aperture_attr_or_default(const gerbv_aperture_t * aperture, 
+                const char * key, const char * dflt)
+{
+    const char * v = _find_value(aperture->attrs, g_intern_string(key));
```
eyal0, line 662: same as above

src/x2attr.c
```diff
@@ -0,0 +1,649 @@
...
+x2attr_get_net_attr(const gerbv_net_t * net, const char * key)
+{
+    return _find_chained_value(net->attrs, g_intern_string(key));
+}
+
+const char * 
+x2attr_get_net_attr_or_default(const gerbv_net_t * net, 
+                const char * key, const char * dflt)
+{
+    const char * v = _find_chained_value(net->attrs, g_intern_string(key));
```
eyal0, line 676: same as above.

How come this one is chained and the rest of them aren't?  Why don't we always use a chained version?

eyal0 (commented):
That was pretty great!  How about we put just this single commit into a PR on its own?  I have reviewed it and and I don't see any problems though I have some suggestions that you could consider.

I think that if we commit this one first, on its own, as it's a big one, maybe we can chip away at this PR little-by-little?

src/export-rs274x.c
```diff
@@ -301,29 +442,28 @@ gerbv_export_rs274x_file_from_image (const gchar *filename, gerbv_image_t *input
...
-					xVal = (long) round(currentNet->start_x * decimal_coeff);
-					yVal = (long) round(currentNet->start_y * decimal_coeff);
-					fprintf(fd, "G01X%07ldY%07ldD02*\n",xVal,yVal);
-				}
-				xVal = (long) round(currentNet->stop_x * decimal_coeff);
-				yVal = (long) round(currentNet->stop_y * decimal_coeff);
-				fprintf(fd, "G01X%07ldY%07ld",xVal,yVal);
+				if ((!insidePolygon) && (currentNet->aperture_state == GERBV_APERTURE_STATE_ON))
+					fprintf(fd, "G01%sD02*\n",_format_pair(&xud, currentNet->start_x, currentNet->start_y, FALSE));
+				fprintf(fd, "G01%s",_format_pair(&xud, currentNet->stop_x, currentNet->stop_y, FALSE));
```
eyal0, line 498: This is a great refactor and I am in favor of it but it could be in its own commit.

src/x2attr.c
```diff
@@ -0,0 +1,649 @@
...
+        s = s->next;
+    }
+    return NULL;
+}
+
+static void
+_populate(gpointer key, gpointer value, gpointer user_data)
+{
+    x2attr_state_t * s = (x2attr_state_t *)user_data;
+    s->t[s->size].key = (const char *)key;
```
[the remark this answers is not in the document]
meantaipan: Yes.  As it mentions in the struct definition, key and value are both interned strings, so their lifetime will persist until GLib is terminated.  Interning makes string compare very fast since just compare pointers, and I'm too lazy to use strdup all over the place.

src/x2attr.c
```diff
@@ -0,0 +1,649 @@
...
+}
+
+static x2attr_state_t *
+_new_from_hash_table(GHashTable * source)
+{
+    x2attr_state_t * s;
+    
+    if (source) {
+        int size = g_hash_table_size(source);
+        s = (x2attr_state_t *)g_malloc0(sizeof(x2attr_state_t) + sizeof(const char *)*2*(size-1));
```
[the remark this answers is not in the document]
meantaipan: Agreed.  Don't know why I didn't do that.

src/x2attr.c
```diff
@@ -0,0 +1,649 @@
...
+_delete(x2attr_state_t * s, const char * key)
+{
+    int i;
+    if (!s)
+        return NULL;
+    i = _find_index(s, key);
+    if (i < 0)
+        return NULL;
+    if (i+1 < s->size)
+        memmove(s->t + i, s->t + (i+1), sizeof(s->t[0])*(s->size-i-1));
```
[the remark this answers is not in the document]
meantaipan: Yes, I may have overestimated the expense of simply resorting.  And looky there: I used sizeof(s->t[0]).  Must have wised up...

src/x2attr.c
meantaipan: The chaining is only used in this case, according to the standard, i.e. for "net" attributes.  In some cases, after looking for a net attribute, also look for aperture attributes.  IIRC, this applies to polygon regions.

rampageservices:
Following up on @eyal0's and @ooxi's earlier requests to break this into smaller pieces — here's an analysis of how the 16 commits decompose into independent, reviewable sub-PRs.

## Commit Grouping Analysis

The commits fall into 7 logical groups with clear dependency boundaries:

### Group 1: RS274-X2 Attribute Infrastructure
**Commits:** `a0f6f2f`, `6354d94`, `71c879d`
**Files:** `src/x2attr.c`, `src/x2attr.h`, `src/gerber.c`, `src/gerbv.h`, `src/export-rs274x.c`

Implements the data model and parser for %TF, %TA, %TO, %TD attribute commands. Modifies the RS274-X exporter to emit attributes. Foundation that everything else builds on. This would also close #362.

### Group 2: IPC-D-356A File Parser
**Commits:** `815ab49`, `606336f`
**Files:** `src/ipcd356a.c`, `src/ipcd356a.h`, `src/main.c`, `src/gerbv.c`

Full parser for IPC-D-356A netlist test data files with CLI options. Addresses #157.
**Depends on:** Group 1 (uses X2 attributes for annotation data)

### Group 3: Geometric Search API
**Commits:** `fc1dc02`, `b6b7554`
**Files:** `src/search.c`, `src/search.h`

Spatial search over rendered geometry — find objects at a given coordinate. Used for hover identification and annotation matching.
**Independent** — pure geometry, no attribute dependency.

### Group 4: File Type Detection Refactor
**Files:** `src/file_type.c`, `src/file_type.h`

Extracts file type detection into its own module. Standalone cleanup.
**Independent.**

### Group 5: IPC-to-Gerber Annotation
**Commits:** `46de144`, `d3ff6c8`
**Files:** `src/gerbv.c` (annotation logic)

Matches IPC-D-356A testpoint data to gerber features by coordinate, annotating objects with net names, refdes, and pin data.
**Depends on:** Groups 1, 2, and 3.

### Group 6: UI Enhancements
**Commits:** `4ac7c47`, `629bbe5`, `cf9f785`, `013aad7`
**Files:** `src/callbacks.c`, `src/draw.c`, `src/render.c`, `src/render.h`, `src/main.c`, `src/main.h`

Mouse hover object info in status bar, net name and refdes highlighting across layers.
**Depends on:** Groups 3 and 5.

### Group 7: Examples, Build System, Man Page
**Commits:** `2d5e33d`, `a09e183`, `2ec9199`
**Files:** `example/rs274-x2/`, `man/gerbv.1.in`, `configure.ac`, `src/Makefile.am`

Should be distributed across the relevant sub-PRs rather than bundled separately.

## Dependency Graph

```
Group 4 (file type detection)     [independent]
Group 3 (geometric search)       [independent]
Group 1 (X2 attributes)          [independent, closes #362]
     │
     ├── Group 2 (IPC-D-356A parser, addresses #157)
     │        │
     │        └──┐
     │           ▼
     └── Group 5 (annotation) ← also depends on Group 3
              │
              ▼
         Group 6 (UI enhancements)
```

## Proposed Sub-PR Order

1. **File type detection refactor** (Group 4) — standalone cleanup
2. **RS274-X2 attribute parsing** (Group 1) — closes #362
3. **Geometric search API** (Group 3) — standalone spatial search
4. **IPC-D-356A file parser** (Group 2) — depends on sub-PR 2, addresses #157
5. **IPC-to-Gerber annotation** (Group 5) — depends on sub-PRs 2, 3, 4
6. **UI: object info and net highlighting** (Group 6) — depends on sub-PRs 3, 5
7. **RS274-X export improvements** (subset of Group 1) — the export refactor that @eyal0 noted could be its own commit

Note: #162 (SEGV on export) is already addressed by PR #379.

I'm filing a tracking issue to coordinate this work and give @meantaipan credit for the original implementation. Each sub-PR would be rebased onto current `main` and submitted independently.

Issue #162, opened 2022-12-18
SEGV on export file as RS274-X


meantaipan:
The following valgrind trace was observed (relative to PR #161 ):

```
steve@charon:~/gerbv/test$ valgrind --trace-children=yes --suppressions=gerbv.supp --error-exitcode=127 --errors-for-leak-kinds=definite --leak-check=full -s --exit-on-first-error=yes --expensive-definedness-checks=yes --keep-stacktraces=alloc-and-free --  gerbv --export=png --window=640x480 --export=rs274x --output=outputs2/example_numpres_numpres.pcb.output_unplated-drill.grb-again outputs2/example_numpres_numpres.pcb.output_unplated-drill.grb
==27813== Memcheck, a memory error detector
==27813== Copyright (C) 2002-2022, and GNU GPL'd, by Julian Seward et al.
==27813== Using Valgrind-3.20.0 and LibVEX; rerun with -h for copyright info
==27813== Command: gerbv --export=png --window=640x480 --export=rs274x --output=outputs2/example_numpres_numpres.pcb.output_unplated-drill.grb-again outputs2/example_numpres_numpres.pcb.output_unplated-drill.grb
==27813== 
==27813== Invalid read of size 8
==27813==    at 0x4E5C227: gerbv_destroy_image (gerb_image.c:192)
==27813==    by 0x429943: main (main.c:1196)
==27813==  Address 0xbe73af0 is 48 bytes inside a block of size 56 free'd
==27813==    at 0x4C31740: free (vg_replace_malloc.c:884)
==27813==    by 0x4E5C23A: gerbv_destroy_image (gerb_image.c:193)
==27813==    by 0x4E5AB68: _export (export-rs274x.c:556)
==27813==    by 0x4E5ABFA: gerbv_export_rs274x_file_from_image (export-rs274x.c:577)
==27813==    by 0x4298D9: main (main.c:1181)
==27813==  Block was alloc'd at
==27813==    at 0x4C33914: calloc (vg_replace_malloc.c:1340)
==27813==    by 0x63D68E0: g_malloc0 (in /lib/x86_64-linux-gnu/libglib-2.0.so.0.4800.2)
==27813==    by 0x4E5C7AE: gerbv_image_return_new_netstate (gerb_image.c:338)
==27813==    by 0x4E6497E: parse_rs274x (gerber.c:1452)
==27813==    by 0x4E61115: gerber_parse_file_segment (gerber.c:268)
[...]

PR #163, opened 2022-12-20
Regression test enhancements: reimport test

meantaipan:
This includes a fix for #162 which was discovered when developing this branch.

Regression tests supports some additional command line options:

```
-i | --reimport        :  Extra export rs274-x, then reimport and check.
-I | --Reimport        :  As above, but fail if images not *identical*

-f | --from-last-err   :  Run from test previously reporting error.
                          If no test failed previously, run from start.

-c | --continue        :  Run from after test previously reporting error.
                          If no test failed previously, exit with message.

-d | --difftool <cmd>  :  Specifies text file diff command.  Will be run as
                            cmd <reference> <output>
                            
-e | --imagetool <cmd> :  Specified image (.png) viewer.  Run as
                            cmd <image>
```

For example, I start off a test run using

`sh run_tests.sh -i -e eog -d meld`

then if anything breaks, run again from that point by adding the `-f` option, or bypass it and run the rest by adding `-c`.

The `-e` and `-d` options make it convenient to examine the results if there are any errors.

Currently, all the tests pass except that reimport does not test the PnP example (because the label text cannot be represented in rs274-x format) and the test-circle-interpolation-1 does not correctly export arcs in quadrant mode.

`-I` (stringent reimport) does not pass, because there are still pixel differences because of resolution conversion etc.  The image comparison process is modified to be a bit more robust for what we want to check:

* Images are output at constant 300dpi.
* Diff is thresholded and eroded - if anything in the erosion, then fail.
* Diff pixels must be against the original image boundary, else fail.

These catch gross errors at about 6-10mil resolution, and also pixel-wide errors that could not be explained by a slight boundary shift or rounding error.

A second rs274-x export file is generated from importing the first, which is compared with the first export output.  These must be exactly equal for the test to pass, since import -> export should be idempotent.


eyal0:
Can we rebase just this change on to gerbv/main so that it doesn't also include all the other stuff?

If you have trouble with it, let me know.  I think that the command that you need to run is like:

```bash
git checkout reimport-test
get rebase HEAD^^ --onto main
```

Something like that?  Check that it looks right and then:

```bash
git push -f
```

ooxi:
@meantaipan no pressure whatsoever, but it would be great if we could merge some of your changes. I suggest going bit by bit, even if it means more work in total. But it also means your work will go into releases of gerbv faster without any pressure for completion :-)

meantaipan:
I'm busy at the moment, but when I get a chance I will take a look at isolating just the fix for #162.

Maybe also the ImageMagick changes to make the regression test image comparisons a bit more robust (not as many false fails).

IIRC, all the other stuff in this fix is fairly dependent on #161, so it doesn't make sense to try to base it directly off the main branch.  I don't mind how long it takes, but have you guys got any estimate for how long the IPC changes will take to get accepted?  I guess there's a lot of code to inspect etc., but just wondering about the normal pace of these things.

ooxi:
@meantaipan thanks for the input! I have not yet looked closer at the IPC-D-356A changes, there have seem to be a couple of changes. I'll add a comment in #161

Issue #171, opened 2023-04-29
Mis-rendering and confusing error message in aperture macro outline primitives with mismatched vertex count


jaseg:
When presented with the following Gerber file, gerbv produces a corrupted rendering result and outputs a misleading error message.

[test.txt](https://github.com/gerbv/gerbv/files/11358913/test.txt)

The file contains two aperture macro definitions. Both use an "outline" primitive. For some reason, in aperture macro outlines, the number of coordinate pairs is listed as a parameter. gerbv internally uses this parameter to allocate memory. When the parameter is mismatched to the actual number of vertices, it produces a cryptic error message on the command line:

```Number of parameters to aperture macro (75) capped to stack capacity (54)```

and produces a corrupted rendering result:

![Screenshot from 2023-04-29 12-19-38](https://user-images.githubusercontent.com/136313/235297820-195d7626-c910-45b2-a0d5-90094e2f7f54.png)

I reproduced this behavior on both gerbv 2.9.5 and 2.9.7-rc1.

## Suggested resolution

1. Detect the mismatch and warn the user with a sepecific warning message such as "Aperture macro outline vertex count does not match parameter count
2. Either:
    a) Use the actual number of parameters instead of the stated vertex count to allocate memory, and render the outline as if the correct count was given (likely to be closest to user intent, and least likely to cause significant mismatch to user intent)
    b) Continue using the stated vertex count to allocate memory, and do not render anything at all when a mismatch is detected (smallest code change, but likely to cause significant mismatch to user intent).

eyal0:
I like your plan to print a warning and then just do what the user probably intended.  Also, it's got to be pretty rare that the parameter count doesn't match the number of parameters!  I can't imagine that many people are writing gerber files by hand and a computer is likely getting this right.

Just do the fix that is easiest and move on.  No one will experience this once  #172 is in.

PR #172, opened 2023-04-29
Bump APERTURE_PARAMETERS_MAX to 10006

jaseg:
Previously, this was set to only 102 and referenced an outdated version of the Gerber file format specification. As of revision 2021.02, the Gerber file format specification lists the maximum vertex count of an aperture macro outline as 5000, and thus the largest possible outline macro might have 10006 parameters.

Spec: https://www.ucamco.com/files/downloads/file_en/416/the-gerber-file-format-specification-revision-2020-09-update_en.pdf#page=62

Section 4.5.1.6:
> The maximum number of vertices is 5000. The purpose of this primitive is to create apertures to
flash pads with special shapes. The purpose is not to create copper pours. Use the region
statement for copper pours; see 4.10

### DIFF
--- a/src/gerbv.h
+++ b/src/gerbv.h
@@ -86,11 +86,12 @@ extern "C" {
 #define APERTURE_MAX 9999
 
 /*
- * Maximum number of aperture parameters is set by the outline aperture
- * macro. There (p. 28) is defined up to 50 points in polygon.
- * So 50 points with x and y plus two for holding extra data gives...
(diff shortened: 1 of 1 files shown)


eyal0:
This is related to #171

eyal0 (approved):

ooxi:
Reintegrated as 1364c29ed048999dc7a060c4b7beeb48e09d9c05, thank you very much!

PR #173, opened 2023-05-01
fix: Using floating point math for arcs.

### LINKED ISSUE #167: Gerbv incorrectly rounds aperture coordinates in fast render mode
In fast render mode, gerbv currently seemingly rounds the screen space coordinates of rendered apertures to the pixel grid based on the top, left corner of the aperture's bounding box. This leads to artifacts where several apertures flashed at the same location are not always rendered concentric with each other when their bounding boxes are different. Ideally, when gerbv needs to round coordinates of an aperture, it should round based on the aperture's origin, which for everything except for aperture macros is the aperture's center.

An example where this rounding error is especially visible is a standard THT pad at low zoom as shown in the example below. At high zoom, the holes are correctly rendered centered on the pads. At low zoom, instead, the holes are rendered shifted to the top, left corner of the pad's bounding box.

Example screenshot at low zoom with the holes rendered shifted:
Screenshot from 2023-04-05 13-18-04

Example screenshot at high zoom with the holes rendered concentric:
Screenshot from 2023-04-05 13-18-10

gerbv version: Version 2.9.5 Compiled on Nov 29 2022 at 23:22:25, from the Archlinux package repository.

eyal0:
Arcs and lines don't always line up when the image is zoomed out because gdk expects integers so we do a lot of integer math.  For example, under a lot of zoom out, we might be looking at arcs circles with a diameter of 5.  When dividing by 2 to compute the radius, this causes a problem.

The solution is to do as much floating-point math as possible and only convert to integer at the very end, when we call the GDK functions to draw.  In testing, this fixed #167.

### DIFF
--- a/src/draw-gdk.c
+++ b/src/draw-gdk.c
@@ -706,11 +706,11 @@ gerbv_gdk_draw_oval(GdkPixmap *pixmap, GdkGC *gc,
 static void
 gerbv_gdk_draw_arc(GdkPixmap *pixmap, GdkGC *gc,
 	       int x, int y,
-	       int width, int height,
+	       gdouble width, gdouble height,
 	       double angle1, double angle2)
 {
-    gint real_x = x - width / 2;
-    gint real_y = y - height / 2;
+    gint real_x = round(x - width / 2);
+    gint real_y = round(y - height / 2);
 
     gdk_draw_arc(pixmap, gc, FALSE, real_x, real_y, width, height,
 		round(64*fmod(angle1, 360)), round(64*(angle2 - angle1)));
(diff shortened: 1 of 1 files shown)


eyal0:
To test, download the files from #167 and then compile and run gerbv like this:

```
make -j 20 && ./src/gerbv ~/Downloads/gerbers-Edge.Cuts.gbr
```

Go to rendering mode "Fast, with XOR".  Then, click on the "-" in the menu bar 6 times.

Before:
image

After:
image

If I zoom out even further than I get some more artifacts but this is at least better than what we had before.  The best solution would be to do all math in floating-point and only convert to integer when necessary.

ooxi:
Thank you! I'll finish the release of 2.9.7 and then merge this

PR #182, opened 2023-07-01
Fix: Translation and Scale are not editable in Edit-layer dialog

kitanokitsune:
gerbv.exe compiled by mingw-w64 with MSYS2 on Windows , in the Edit-layer dialog that opens by [Layer]-[Edit layer], numerical values cannot be entered into the edit box from the keyboard.

According to https://stackoverflow.com/questions/58318673/gtk-window-is-active-not-working-as-expected , the event handler which handles "focus-in-event" should return 0 or 1:
+ If the event handler returns 1 (TRUE), the behavior is handled by that event handler.
+ If the event handler returns 0 (FALSE), the behavior is GTK's default behavior.

Yet *focus_in_event_callback* does not return a value (void).

### DIFF
--- a/src/interface.c
+++ b/src/interface.c
@@ -2303,11 +2303,11 @@ interface_show_alert_dialog (gchar *primaryText, gchar *secondaryText,
 
 static int focused_widget_num = 0;
 
-int
+gboolean
 focus_in_event_callback (int *widget_num)
 {
 	focused_widget_num = *widget_num;
-	return 0;
+	return FALSE;
 }
 
 void
--- a/src/interface.c
+++ b/src/interface.c
@@ -2303,10 +2303,11 @@ interface_show_alert_dialog (gchar *primaryText, gchar *secondaryText,
 
 static int focused_widget_num = 0;
 
-void
+int
 focus_in_event_callback (int *widget_num)
 {
 	focused_widget_num = *widget_num;
+	return 0;
 }
 
 void


ooxi:
I cannot reproduce the issue, it's fine entering a value from the keyboard into layer translation and scale.

Nevertheless, I'm fine with changing the callback signature. But I cannot find any reference to it accepting `int` as return?

According to [gtkd.org](https://api.gtkd.org/gobject.c.types.GCallback.html) it should have a `void` return type?

kitanokitsune:
Thank you for comment.

> Nevertheless, I'm fine with changing the callback signature. But I cannot find any reference to it accepting `int` as return?
> 
> According to [gtkd.org](https://api.gtkd.org/gobject.c.types.GCallback.html) it should have a `void` return type?

A certain source file on the web used `int` type, so I copied it.
However, in response to your comment, I have investigated again.
And I found some samples ([ [_focus_in_cb](https://github.com/ibus/ibus/blob/main/client/gtk2/ibusimcontext.c#L288) ], [ [GUI_ed1_cb1, GUI_ed1_cb2, GUI_ed1_cb3](https://gcad3d.org/gui_gtk2/gtk_ed1.c) ], [ [Drag and drop example](https://zetcode.com/gui/gtk2/gtkevents/) ]) which callbacks do not return `void`.

Indeed, the link you posted above also said the following:

_This doesn't mean that all callback functions must take no parameters and return void.
The required signature of a callback function is determined by the context in which is used (e.g.  the signal to which it is connected)._

This means that the signature of the callback function seems to be context dependent.
In our case, the ["focus-in-event"](https://developer-old.gnome.org/gtk2/stable/GtkWidget.html#GtkWidget-focus-in-event) signal handler returns `gboolean` type. Therefore, it seems to me that the relevant code should have been modified as follows
```
gboolean
focus_in_event_callback (int *widget_num)
{
	focused_widget_num = *widget_num;
	return FALSE;
}
```

ooxi:
Thank you @kitanokitsune for all your work! It would be great if we can merge our efforts :-)

PR #183, opened 2023-07-01
Japanese Language File

kitanokitsune:
I made a Japanese language file.
Please merge it!


ooxi:
Thank you very much! I see not all strings are translated, will they remain in English (which is ok) or not be displayed?

kitanokitsune:
Thank you for merging it!

> I see not all strings are translated, will they remain in English (which is ok) or not be displayed?

This is due to Windows character encoding reason.
On Windows, **UTF-8** encoded strings are used in GTK internal, however, text console (stdout/stderr) uses system encoding (_e.g._ **cp932** on Japanese Windows). Therefore, strings containing multibyte characters need to be left untranslated because they will be garbled when output to the console.

Issue #107, opened 2022-04-19
gerbv should be able to detect when a project file is passed as argument without option --project="..."


Dromantor:
From what I could see gerbv currently expects that each argument that is not an option is a valid gerber/drill file. The problem is that if you just configure to open .gvp with gerbv it fails to do so because your OS (I'm on Windows btw.) doesn't know about the needed --project="..." option. You currently have to manually modify the command line used to open your project file, which may or may not be easy depending on your OS.

Another problem with projects: The file `init.scm` that seems to be required for opening a project isn't packaged with the builds.

ooxi:
This does make sense from a usability point of view. However, the argument parsing code is quite elaborated. I would be happy to mentor such a patch however, if you want to create a PR.

Could you post an error message regarding `init.scm`? Indeed, it is not bundled and might be a simple fix.

Dromantor:
I'll try to come up with something when I got some spare time.

And regarding the `init.scm` error: I guess you already found the relevant error message yourself :)

ooxi:
@Dromantor the `init.scm` issue #111 was resolved by pull request #112 and will be included in the next release. You can use [snapshot builds](https://gerbv.github.io/#download) until then.

ooxi:
Fixed by PR #184

PR #184, opened 2023-07-02
Allow 'gerbv foo.gvp' to act like 'gerbv -p foo.gvp' was specified

kitanokitsune:
Allow .gvp file to open without -p option.
This is from [#76 Allow 'gerbv foo.gvp' to act like 'gerbv -p foo.gvp' was specified](https://sourceforge.net/p/gerbv/patches/76/) created by Greg Ercolano.

This patch resolves issue [#107](https://github.com/gerbv/gerbv/issues/107).

### DIFF
--- a/src/gerbv.c
+++ b/src/gerbv.c
@@ -1114,3 +1114,11 @@ gerbv_transform_coord_for_image(double *x, double *y,
 	return 0;
 }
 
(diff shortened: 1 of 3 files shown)


ooxi:
Thank you very much!

PR #186, opened 2023-07-06
Increase maximum zooming level

### LINKED ISSUE #185: Zoom max level / GERBV_SCALE_MAX (and Windows build)
Hi, 

In 2018, for a work colleague who wanted to make precision measurements, I recompiled under Windows a version of gerbv (2.6A) with an increased zoom level. 
At the time I modified the #define GERBV_SCALE_MAX from 3000 to 40000 in gerbv.h (https://github.com/gerbv/gerbv/blob/main/src/gerbv.h#L96C12-L96C12)

Now my colleague would like a more up-to-date version but with the same increased zoom limit.
Unfortunately, the Windows compilation doesn't seem to be up to date, so this ticket https://github.com/gerbv/gerbv/issues/19#issuecomment-931499319 seems to say that the "win32" directory is completely obsolete?

So, in order of simplicity for me: 
1) Can you change the max zoom limit in the official version?
2) If you can't, is there any up-to-date documentation for compiling a Windows binary?

Thanks, 
Stéphane

ooxi:
Fixes #185


eyal0:
Seems fine!

PR #187, opened 2023-07-06
Port to GTK3

LemonBoy:
Initial port to GTK3, looks functional enough at first glance.

There's still a lot of deprecation warnings, but the code builds and the UI works.


ooxi:
This is a huge contribution, thanks for the work you poured into this! Besides removing the GDK renderer, are there other functions removed?

ooxi:
I got the CI working for Fedora! The `configure.ac` script seems to have a problem with dxflib detection on Ubuntu.

The GUI looks really rejuvenated. However, it's much harder to spot differences in the XOR renderer. Moreover, is there a way to make the Cairo renderer faster? I have trouble viewing big gerber files with Fast already, but cairo is unusable.

Nevertheless, great work!

ooxi:
I got the CI working for Fedora! The `configure.ac` script seems to have a problem with dxflib detection on Ubuntu.

The GUI looks really rejuvenated. However, it's much harder to spot differences in the XOR renderer. Moreover, is there a way to make the Cairo renderer faster? I have trouble viewing big gerber files with Fast already, but cairo is unusable.

Nevertheless, great work!

LemonBoy:
> Besides removing the GDK renderer, are there other functions removed?

Nope, some GUI elements may be slightly off but haven't noticed anything too strange.

> However, it's much harder to spot differences in the XOR renderer.

Is it? Cairo has no bitwise xor operator so I had to use the difference one, but the result look quite similar to me. The only notable difference is that the alpha value is now taken into account, while the old XOR renderer didn't. I'll try to disable the alpha blending to see if the results improve.

> Moreover, is there a way to make the Cairo renderer faster? 

No idea, haven't run a profiler on it yet. Hopefully the answer is yes :)

ooxi:
> Nope, some GUI elements may be slightly off but haven't noticed anything too strange.

🥳 didn't notice other things in preliminary testing, too


> Is it? Cairo has no bitwise xor operator so I had to use the difference one, but the result look quite similar to me.

the colours seem to be alpha blended. I am sure there is a way to just change the palette and get a visually more distinct result


My preliminary course of action would be to include this as a `feature/gtk3` branch and rebase on `main` for new changes, until we will reach a defined set of conditions for a big new release.

@eyal0 / @spe-ciellt: In my opinion, moving to GTK3 is the right way forward, even if we lose features, since GTK2 is EOL. This is a change, worthy at least a 2.10.x, in my opinion even a 3.x version bump.

We could maintain a 2.9.x branches as long as changes can be merged.

Do you have input on this?

spe-ciellt:
@ooxi Sorry for late reply.
@LemonBoy Thank you for this major improvement and update.

I think this is really great and significant update! I think the best way forward is as you suggest to have it not merged into main yet, but to have it on a feature branch. But when merged it is probably worth a 3.x update.

Rendering speed have always been a problem for gerbv, hence the different renders. As soon as I tried to retire the fast render complaints about rendering speed poured in.

I think the lack of rendering speed for gerbv comes from a fundamental "design decision". We are blitting pixels onto the screen, while I think (wildly guessing here) is that for fast rendering you need to render polygons or something similar.
And the image is "rerendered" every pan/zoom. But I think that rendering speed should not limit us to go to 2023. ;)

@ooxi @eyal0 Thank you for all the hard work you put into this.

ooxi:
Thank you for your feedback @spe-ciellt! I have thought about this some more and I'm really torn between keeping this in a feature branch for 3.x release and merging into `main` behind a `GTK3` define.

Using a define will lower the maintenance burden for the patch set and give distributions an opportunity to enable whenever they want. However, it will also increase the risk of having to support two GUI toolkits for an extended period of time.

@LemonBoy would you even be willing to put in the work to refactor this patchset into a GTK2/GTK3 define? This is a huge ask, so don't feel pressured. But it will also increase the velocity of releasing this work into distributions.

eyal0:
@ooxi @LemonBoy Regarding the `#ifdef`, this is a feature of the unix `diff` tool.  If you provide `--ifdef=GTK3_SUPPORT` for example, it'll make all the `#ifdef` lines for you.

I dunno if this is necessarily the correct direction but, just so that you know, this option exists!

It's best to limit the number of `#ifdef` for readability so I would suggest that we use wrappers where possible.

For example, there are many places in this pull request where [`gtk_hbox_new (FALSE, 4)`](https://docs.gtk.org/gtk3/ctor.HBox.new.html) is changed to [`gtk_box_new (GTK_ORIENTATION_HORIZONTAL, 4)`](https://docs.gtk.org/gtk3/ctor.HBox.new.html).

So rather than have many `#ifdef` in the code, you can change all those old `gtk_hbox_new(FALSE, 4)` to `gerb_gtk_box_new(GTK_ORIENTATION_HORIZONTAL, 4)`.  And then add a function like this:

```c
GtkWidget gerb_gtk_box_new(GtkOrientation orientation, int spacing) {
  #ifdef GTK3
    return gtk_box_new(orientation, spacing);
  #else
    if (orientation == GTK_ORIENTATION_HORIZONTAL) {
      return gtk_hbox_new(FALSE, spacing);
    } else {
      return gtk_vbox_new(FALSE, spacing);
    }
  #endif
}
```

This way all the calls within the code look the same, they are all `gerb_gtk_box_new`.  And only within `gerb_gtk_box_new` do you have any of the `ifdef` stuff.

I would even take it one step further and split the commit into two commits.  In the first commit, you just convert all the `gtk_hbox_new` stuff into `gerb_gtk_box_new` and have `gerb_gtk_box_new` look like this:

```c
GtkWidget gerb_gtk_box_new(GtkOrientation orientation, int spacing) {
  if (orientation == GTK_ORIENTATION_HORIZONTAL) {
    return gtk_hbox_new(FALSE, spacing);
  } else {
    return gtk_vbox_new(FALSE, spacing);
  }
}
```

Basically, that commit will be just a refactoring that doesn't affect the operation of the code at all.  Such refactorings are easy to examine by eye and confirm that they are done correctly.  So if this one is a big commit, no problem.

And then you follow that commit with the commit that changes the internals of `gerb_gtk_box_new`.  This will be the complex commit where we are adding support for GTK3 and potentially many things could break.  *However*, because of the refactoring from the commit before it, it'll be a pretty small commit.

So instead of one big, complex commit, you get one big commit that is simple and one small commit that is complex.

Anyway, I've done this kind of thing a few times in my life so I'm happy to help out.  The first step would be to get all this GTK3 stuff working and passing CI.  And after that, then we can look into how to re-arrange the code so that it's in nice, manageable commits as much as possible.

Another advantage of doing it in the multiple commits like I suggest is that, should something go haywire, it's easy to rollback the problematic changes without rolling back the changes that worked.

I hope that this all makes sense!  I have almost zero GTK experience but if I can be of help with any sticky problem, let me know.

LemonBoy:
> would you even be willing to put in the work to refactor this patchset into a GTK2/GTK3 define?

Not really.
My two cents. I feel time would be better spent rewriting the renderer to be faster/snappier, older gerbv versions will always be available for users _needing_ the GTK2 version.

Salamandar:
I can contribute a move from Configure.ac to Meson build system to fix some issues @LemonBoy

LemonBoy:
A revamped build system either using CMake or Meson and some kind of automatic formatter (clang-format) would really help further work.

eyal0:
> A revamped build system either using CMake or Meson and some kind of automatic formatter (clang-format) would really help further work.

Gerbv is written in c, not even c++. Introducing a dependency on python might be inappropriate. Meson uses python.

For a c-only project, cmake is probably the most appropriate. Right now we're using autotools and cmake seems like a sideways move more than an upgrade.

You can add clang-format into the CI right now if you want. Make a PR that enforces clang format in CI and we can take a look. Personally, I hate the mix of tabs and spaces and I'd welcome switching over to all spaces. (I've never in my life seen a project that uses tabs and spaces correctly.)

Salamandar:
> Gerbv is written in c, not even c++. Introducing a dependency on python might be inappropriate. Meson uses python.

Don't forget this is only a build dependency. Also there are working C implementations of Meson.

And, in some ways, the autotools dependency is even harder to install and set up than a simple python package (don't get me started on autoconf versions incompatibility, m4, etc).

> For a c-only project, cmake is probably the most appropriate. Right now we're using autotools and cmake seems like a sideways move more than an upgrade.

I don't really like Cmake (although i'm an ex-Cmake fan) but Meson is really nice to use, and is really POSIX/freedesktop friendly.
Also most gtk/gnome projects are now using Meson, and they all are C projects.

The fact that Meson's official implementation is written in Python shouldn't impact the projects build with Meson. In fact I ported a lot of softwares (including Gimp) and most of the time i'm trying to not change a single line of the source code (mostly to have a period when Autotools and Meson are both provided).

ooxi:
> My two cents. I feel time would be better spent rewriting the renderer to be faster/snappier, older gerbv versions will always be available for users needing the GTK2 version.

I completely agree! Nevertheless, I also see the reality of nobody stepping up to write a faster Cairo based renderer. Your contributions are hugely valuable, we just have to decide how to best digest them.


> I can contribute a move from Configure.ac to Meson build system to fix some issues @LemonBoy

While I personally am more familiar with CMake, I welcome a move to each of the modern build systems like Bazel, CMake and Meson. Being a fanboy of jpakkane I do support your enthusiasm :-)

But this is completely orphogonal of moving from GTK2 to GKT3

Patronics:
Any news on this? It seems that gerbv is now [listed as deprecated in homebrew](https://formulae.brew.sh/formula/gerbv) due to the dependency on gtk2. It'll be fully disabled in homebrew if not fixed by 2027-08-28.

Salamandar:
> Any news on this? It seems that gerbv is now [listed as deprecated in homebrew](https://formulae.brew.sh/formula/gerbv) due to the dependency on gtk2. It'll be fully disabled in homebrew if not fixed by 2027-08-28.

Well i'm all for helping, i have a branch somewhere for a Meson build.

PR #192, opened 2023-07-12
fix: Out-of-bounds memory access in strtok.

### LINKED ISSUE #191: Denial of Service in Gerbv
### Summary

Error or warning while processing an included image may result in a crash of Gerbv.

### CWEs

- CWE-824: Access of uninitialized pointer

### Details

The Gerber RS-274X format defines the `IF` command (`INCLUDE FILE`, for referencing an external file) in its [User's Guide](https://www2.pv.infn.it/~servel/rc_images/gerber_rs274xrevd_e.pdf) (page 28). A valid line using this command respects the format below:

```
%IF<filename>.<extension>*%
```

The implementation of this functionality leverages two utilities:
- A `gerb_file_t` structure wrapping up details about an opened file, such as the file descriptor, filename, and data to be read; and
- A `gerb_fopen` function (from `gerb_file.c`) that takes a filename and returns a (partially-initialized) `gerb_fopen` structure;

The relevant execution flow in the context of this issue is as follows:
1. A `gerbv_open_image` function (from `gerbv.c`) is initially called in `main.c` through a proxy function called `main_open_project_from_filename`.
2. The `gerbv_open_image` internally calls `gerb_fopen` and stores the filename as a structural member in `fd->filename = g_strdup(filename)`.
3. In a call to `gerber_is_rs274x_p`, the file is checked to respect the RS-274X format.
4. If it respects the format, `parse_gerb` is called.
5. `parse_gerb` calls `gerber_parse_file_segment`.
6. If any command having a `%` is detected, `parse_rs274x` is called.
7. If the command is detected to be `IF`, then the file is read with `gerb_fopen`.
8. `gerber_parse_file_segment` is called again, this time having as argument a filename stored inside the initially-read file.

The issue resides in the fact that, compared to the second step, the eighth one does not set the filename. When the latter is accessed in further processing, an invalid address is dereferenced, which results in a crash of the program.

A code path achieving this is when the included files contain an invalid M-code (for example, `M09*`). Usually, despite the invalid code, the execution should continue without processing the erroneous information, but Gerbv executes the following in its `parse_M_code`:

```
gerbv_stats_printf(stats->error_list, GERBV_MESSAGE_ERROR, -1,
		_("Encountered unknown M%02d code at line %ld in file \"%s\""),
		op_int, *line_num_p, fd->filename);
```

As it can be observed, the `fd->filename` is accessed in a logging call, so the execution will crash instead of raising a warning and continuing. This will not happen if the random value found on heap and implicitly assigned to `fd->filename` is `NULL`. Most compilers have a check against this scenario, and will only print `(null)`.

### Affected Releases

The functionality of processing `IF` commands was introduced in `a5b8484`, which corresponds to `v2.4.0`.

### PoC

For reproducing this, Cisco's `parse_aperture_strtok.1.poc` from CVE-2021-40401 was used as the file provided as an input parameter. The content of `parse_aperture_strtok.2.poc` was replaced with `M09*`.

As the issue depends on the random value of the unitialized `fd->filename`, it was reproduced by launching `gdb`, setting a breakpoint on `gerb_fopen`, and setting `fd->filename` to the value `0x0101010101` (a random value that could be found on the heap). Gerbv will generate a crash as follows:

```bash
$ gdb ./gerbv
[...]
(gdb) c
Continuing.

Program received signal SIGSEGV, Segmentation fault.
0x0000fffff7376590 in ?? () from /lib/aarch64-linux-gnu/libc.so.6
(gdb) bt
#0  0x0000fffff7376590 in  () at /lib/aarch64-linux-gnu/libc.so.6
#1  0x0000fffff734b848 in  () at /lib/aarch64-linux-gnu/libc.so.6
#2  0x0000fffff735cd78 in  () at /lib/aarch64-linux-gnu/libc.so.6
#3  0x0000fffff73cfa3c in __vasprintf_chk () at /lib/aarch64-linux-gnu/libc.so.6
#4  0x0000fffff7648064 in g_vasprintf () at /usr/lib/aarch64-linux-gnu/libglib-2.0.so.0
#5  0x0000fffff761da88 in g_strdup_vprintf () at /usr/lib/aarch64-linux-gnu/libglib-2.0.so.0
#6  0x0000fffff7fa3d04 in gerbv_stats_printf (list=0xaaaaaab2ab20, type=GERBV_MESSAGE_ERROR, layer=-1, text=0xfffff7fb5830 "Encountered unknown M%02d code at line %ld in file \"%s\"") at gerb_stats.c:242
#7  0x0000fffff7fa7088 in parse_M_code (fd=0xaaaaaab2ced0, image=0xaaaaaab16fc0, line_num_p=0xffffffffe0f0) at gerber.c:1183
#8  0x0000fffff7fa487c in gerber_parse_file_segment
    (levelOfRecursion=2, image=0xaaaaaab16fc0, state=0xaaaaaab16f60, curr_net=0xaaaaaab2af60, stats=0xaaaaaab2aa70, fd=0xaaaaaab2ced0, directoryPath=0xaaaaaab14a60 "/home/gerbv/triggers")
    at gerber.c:172
#9  0x0000fffff7fa7f40 in parse_rs274x
    (levelOfRecursion=1, fd=0xaaaaaab149a0, image=0xaaaaaab16fc0, state=0xaaaaaab16f60, curr_net=0xaaaaaab2af60, stats=0xaaaaaab2aa70, directoryPath=0xaaaaaab14a60 "/home/gerbv/triggers", line_num_p=0xffffffffe350) at gerber.c:1459
#10 0x0000fffff7fa4bec in gerber_parse_file_segment
    (levelOfRecursion=1, image=0xaaaaaab16fc0, state=0xaaaaaab16f60, curr_net=0xaaaaaab2af60, stats=0xaaaaaab2aa70, fd=0xaaaaaab149a0, directoryPath=0xaaaaaab14a60 "/home/gerbv/triggers")
    at gerber.c:245
#11 0x0000fffff7fa60a8 in parse_gerb (fd=0xaaaaaab149a0, directoryPath=0xaaaaaab14a60 "/home/gerbv/triggers") at gerber.c:770
#12 0x0000fffff7facca0 in gerbv_open_image
    (gerbvProject=0xaaaaaab14940, filename=0xffffffffed6e "/home/gerbv/triggers/CVE-2021-40401.1.poc", idx=0, reload=0, fattr=0x0, n_fattr=0, forceLoadFile=1) at gerbv.c:527
#13 0x0000fffff7fabf44 in gerbv_open_layer_from_filename_with_color
    (gerbvProject=0xaaaaaab14940, filename=0xffffffffed6e "/home/gerbv/triggers/CVE-2021-40401.1.poc", red=29555, green=29555, blue=57054, alpha=45489) at gerbv.c:249
#14 0x0000aaaaaaacb334 in main (argc=6, argv=0xffffffffe8e8) at main.c:937
```

### Impact

Gerbv can be crashed when processing files.

### Suggestions for Patching

The filename can be attached to the `gerb_file_t` structure in `gerb_fopen` to avoid invalid dereferencing.

eyal0:
This fixes #191


ooxi:
Thank you for the testcase! I think the best way forward is to set a default value for `fd->filename` in `gerb_fopen`. I was considering `strdup`ing the `filename` argument, but that might lead to even more memory leaks.

Thus I'll try setting it to `nullptr` first.

ooxi:
I have tried both `nullptr` and `""` as default initialization. In both cases the *Conditional jump or move depends on uninitialised value* message goes away, but valgrind does not pass, either.

However, I cannot make out the cause of the error. Do you see an error: https://github.com/ooxi/gerbv/actions/runs/5545198229/jobs/10123767309?

eyal0:
```
==32056== 28 bytes in 1 blocks are definitely lost in loss record 110 of 296
==32056==    at 0x483C855: malloc (vg_replace_malloc.c:393)
==32056==    by 0x506BE98: g_malloc (in /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0.6400.6)
==32056==    by 0x4865BD6: gerb_fgetstring (gerb_file.c:219)
==32056==    by 0x486BFC1: parse_rs274x (gerber.c:1450)
==32056==    by 0x486A68F: gerber_parse_file_segment (gerber.c:245)
==32056==    by 0x486E507: parse_gerb (gerber.c:770)
==32056==    by 0x486F148: gerbv_open_image (gerbv.c:571)
==32056==    by 0x486F59B: gerbv_open_layer_from_filename_with_color (gerbv.c:249)
==32056==    by 0x11B452: main (main.c:941)
```

@ooxi Setting it to `nullptr` or `""` fixed the problem of trying to read bad memory.  However, it exposed a different problem, which I also encountered, same as you!  It's saying that there is memory that was allocated but never freed, and it shows the stack trace to that allocation.  Along that path, I can see `parse_rs274x (gerber.c:1450)`.  In the code, it's here: https://github.com/gerbv/gerbv/blob/caa6560d5d683f827c672fd5e380f89a8ef632b6/src/gerber.c#L1450-L1480

Looking into the code for that call to `gerb_fgetstring`, I see: https://github.com/gerbv/gerbv/blob/caa6560d5d683f827c672fd5e380f89a8ef632b6/src/gerb_file.c#L196-L224

There's a `g_malloc` in there with no corresponding free.  So it needs to be free'd!  We've never caught this before in valgrind because we have never covered this part of the code with valgrind.  :-(  So that's where the memory is getting lost.  You can see that the size lost is 28 bytes and the file name that @iosifache chose was `parse_aperture_strtok.2.poc` and that's 27 letters.  Add one for the terminating null and it's 28, same as the `malloc` size.

ooxi (approved):
Thank you very much!

PR #194, opened 2023-07-13
Fedora 36

eyal0:


spe-ciellt:
The current develop branch currently uses fedora 43.

PR #197, opened 2023-08-19
enable manual workflow initiation

henrygab:
This just allows folks to manually initiate the CI action on their own forks, without needing to generate a PR.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -8,7 +8,7 @@ on:
       - main
   pull_request:
     branches: [ main ]
-
+  workflow_dispatch:
 
 jobs:
   ci:


eyal0:
I think that the colon after workflow_dispatch is unnecessary but there's no harm in it!

Looks good to me

ooxi:
Thank you for this PR! Please be aware that GitHub will update a PR if you add more commits to the same branch, making the PR harder to merge.

I will thus merge da3999948b6deafeb5b399d67df049799f63532e manually

henrygab:
Thanks for catching the extra commit ... went to the wrong branch.  I've force-pushed back to da39999 to track this back to here.  Thanks!

henrygab:
NOTE: Please consider merge of PR #199 first.

_(else it will need adjustments to the `.git-blame-ignore-revs`)_

Thanks!

Issue #195, opened 2023-08-03
Drill files get scaled incorrectly


henrygab:
May be related to #174 ??

* PCB Viewer in KiCad 6.0 shows everything aligned properly.
* JLCPCB.com online gerber viewer shows everything aligned properly.
* tracespace.io/view online gerber viewer shows everything aligned properly.

However, using GerbV will show the drills at the wrong location (wrong scale?).
Here's an ultra-tiny repro:

[GerbV_Bug.zip](https://github.com/gerbv/gerbv/files/12245790/GerbV_Bug.zip)

It appears by removing ALL the commented lines at the top of the drill files, the drills are then properly parsed / shown.

This is why I thought it might be related to #174.  (I'm opening a separate issue until it's determined to be the same.)

ooxi:
Thank you for the report! This seems to be related to #174 and provides a good test case for a fix!

While I currently do not have the capacity to provide a fix, I'm happy to provide guidance

henrygab:
> While I currently do not have the capacity to provide a fix, I'm happy to provide guidance

I have reviewed enough of the code to make a recommendation, and can likely provide a PR with fixes.  I'd appreciate your thoughts....


<hr/>
<details><summary><b>Recommendation (click to expand/collapse)</b></summary><p>

(All changes appear to reside in `src/drill.c`.)

* Define new struct `number_format_options_t`:
  * `uint8_t leading_digits` -- stores only the count of leading digits
  * `uint8_t trailing_digits` -- stores only the count of trailing digits
  * `gerbv_unit_t` -- stores the units (inches, mm, unspecified)
  * TBD ... if code actually requires `number_fmt_t` (e.g., to output a pre-defined format using short notation), then can also add that as a field of this new structure ...
* In existing `drill_state_t` structures:
  * TBD ... should existing `gerbv_unit_t unit` field be removed?
  * Replace three existing `number_fmt_t` fields with new `number_format_options_t` fields
* Adjust code accordingly (some comments already added above)
  * TBD ... add helper and internal consistency validation functions for `number_format_options_t`
* In addition, improve README file format requirements adherence / consistency 
  * convert all tabs into spaces
  * apply braces consistently
  * try to declare variables according to their needed scope (less mental load)
  * use more descriptive variable names where possible

</p>
</details><hr/>

[...]

PR #199, opened 2023-08-20
Clang format

henrygab:
Per @ooxi 's [comment](https://github.com/gerbv/gerbv/issues/195#issuecomment-1684941340),
I have prepared this PR to perform 100% automated cleanup of the formatting using `clang-format`.

Next up will be to enable a linter (via CI / github actions) that ensures code continues to conform.  Getting this onto the radar now, in case additional questions arise.

NOTE: builds off PR #197 "enable manual workflow initiation".


eyal0:
I'll suggest something, that you can choose to do in this PR or another one:

`git blame` is now going to blame you for a lot of changes in the repo because of all the changes that clang-format made.  To prevent this, `git blame` has a feature called `--ignore-revs-file`.  The documentation for it is:

```
--ignore-revs-file <file>
Ignore revisions listed in file, which must be in the same format as an fsck.skipList. This option may be repeated, and these files will be processed after any files specified with the blame.ignoreRevsFile config option. An empty file name, "", will clear the list of revs from previously processed files.
```

I suggest that you create a file called `.git-blame-ignore-revs` in the root of this repo.  In that file, put the commit SHA that cause all the whitespace changes.  So the file has just this:

```
bb965cbc3378220b1505f38b62639d124b41e8d5
```

And then users that want to ignore this rev can type:

`git config --local blame.ignoreRevsFile .git-blame-ignore-revs`

Also, this file name is a special file name that is recognized by github.  GitHub will automatically offer to ignore commits in that file, as described here: https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view

I haven't tested any of this before so you'll have to check it out but it seems straightforward.

ooxi:
Thank you very much, we will merge this and do a patch release with all the changes. One request: please do a PR rebased on `main` containing only the clang-format change.

henrygab:
**_I believe this fully matches the requested changes.  If not, please let me know._**

* Recreated branch from main (19b8baf, aka tag: v2.10.0-rc.1)
* Added `.clang-format` file as separate commit (blame remains on me for that file)
* Applied formatting via `clang-format -i -style=file src/*.h src/*.c src/*.cpp`
* Created `.git-blame-ignore-revs` file, per your most excellent suggestion!
* Force-pushed my new resulting branch over the old one ... thus allowing the PR to track its evolution, while keeping your merge options open.

**_NOTE:_** Unless this is the next PR to be merged, I think the commit hash may change.  Similarly, I am presuming this PR will be a ff-only merge from 19b8baf, else (again) the hash in the new ignore file will need to adjusted manually during merge.


P.S. That's information about the `.git-blame-ignore-revs` is **AWESOME**!  Such a useful feature;  I love learning new tips and tricks ... **_thank you_**!

henrygab:
> What's the right way to merge this so that we don't have the SHA change? Maybe "Create a merge commit" and not "Rebase and merge" nor "Squash"?

Standard merge uses fast-forward, and would result in the desired outcome.
From the command line in main branch, you can restrict a merge to require fast-forward:

```
git switch main
git remote add henrygab https://github.com/henrygab/gerbv.git
git merge --ff-only remotes/henrygab/clang_format
```

If the standard merge isn't shown (but "rebase and merge" is shown), that would be the second-best option.  Basically, complete the rebase, then add a commit to add the updated commit hash.  Not terribly expensive, but adds steps.

FYI, I'm pretty sure there's a way to merge directly from a specific commit (in this case, bc9a6990dfd576d55738fa75bc520cca841ba127) rather than branch name.   That said, I've never tried to do that, while I use `--ff-only` regularly(*) with branch names ... so that's what I showed above.


<details><summary><B>*</B></summary><p/>

For example, I may rebase'ing for "leaf" branches (branches that aren't themselves an integration branch).
This helps keep a linear history (useful for versioning), while retaining multiple smaller commits (useful for tracking bug introduction / git blame / fewer and smaller merge conflicts).

As rebase has other detrimental effects for non-leaf branches, I understand when other folk have different opinions on the cost vs. benefits of rebasing.

</details>

henrygab:
Let me rebase again ... looks like I missed a commit from ~13 hours ago.

henrygab:
OK, done.  Rebased against current main HEAD, which is now 76ec1da56e37de3e069426303aec1857e31dcd26.  This changed the commit hash for the `.git-blame-ignore-revs` file, as expected.  I updated so it should be a simple PR.

ooxi:
Thank you very much @henrygab! It would be great, if we could add a GitHub Action check which fails if `.clang-format` is not respected.

I want to do a 2.10.1 soon with just these changes :-)

eyal0:
I can help with the clang format check if needed. It's pretty easy to do in CI. You might need to install clang. Just do it in the run that does the coverage tests, we don't need a clang check in both CI runs.

henrygab:
> I can help with the clang format check if needed. It's pretty easy to do in CI. You might need to install clang. Just do it in the run that does the coverage tests, we don't need a clang check in both CI runs.

Yes, I would greatly appreciate your handling the CI check!   I have added CI to some depots, but your knowledge of the current setup / prerequisites / etc. makes it highly more likely you will get it right the first time.  :)  Thank you, @eyal0 !

PR #201, opened 2023-08-23
ci: Run clang-format.

### LINKED ISSUE #200: ci: Run clang-format
> I can help with the clang format check if needed. It's pretty easy to do in CI. You might need to install clang. Just do it in the run that does the coverage tests, we don't need a clang check in both CI runs.

Yes, I would greatly appreciate your handling the CI check!   I have added CI to some depots, but your knowledge of the current setup / prerequisites / etc. makes it highly more likely you will get it right the first time.  :)  Thank you, @eyal0 !

_Originally posted by @henrygab in https://github.com/gerbv/gerbv/issues/199#issuecomment-1688920917_

eyal0:
This fixes #200

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -30,9 +30,27 @@ jobs:
     - name: Set a local install path
       run: echo "LOCAL_INSTALL_PATH=${HOME}/.local" >> $GITHUB_ENV
 
+    - name: Update apt sources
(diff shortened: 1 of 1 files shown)


eyal0:
In my opinion, this is ready to go.  @ooxi , take a look at merge it if you're pleased.

eyal0:
We need to put clang-format into the pre-commit hook, too.

ooxi:
This is awesome! Thank you @eyal0

PR #203, opened 2023-08-23
Clang format precommit hook

### LINKED ISSUE #202: build: Put clang-format into the pre-commit hook.
We need to put clang-format into the pre-commit hook, too.

_Originally posted by @eyal0 in https://github.com/gerbv/gerbv/issues/201#issuecomment-1690260110_

eyal0:
This uses the `.clang-format` file to format any changed c or c++
code.

It uses whichever version of `clang-format` that can be found on the
compute that is newer than the specified major version.  The default
is to look for version 14 or later.  If a suitable `clang-format` is
not found, the program fails.

Only changes that are in the index (ie, that have been added by `git
add`) are formatted.  Changes in the worktree that aren't yet added to
the index are not formatted.

Any formatting changes that are found necessary are applied to both
the index and the current worktree.  This might fail if the index and
the worktree have diverged so significantly that the changes can't be
patched to the worktree.  This ought to be very rare.

A `-n` (equivalently, `-dry_run`) flag can be patched to the program
to cause the program to make no changes to any files.  Only the diff
will be printed out.  This diff can be applied by redirecting the
output of the dry-run to `patch -p1`.

This fixes #202

### DIFF
--- a/clang-format-cache.py
+++ b/clang-format-cache.py
@@ -1,0 +1,126 @@
+#!/usr/bin/env python
+
+'''clang format the files in the index (aka the staged files).'''
+
+import os
+import re
+import subprocess
+import sys
+import fnmatch
+import difflib
+import argparse
+
+
+def find_executable(filename_regex):
+  '''Find executables by regex and yield them all.'''
+  for path in os.environ["PATH"].split(":"):
+    if not os.path.isdir(path):
+      continue
+    for filename in os.listdir(path):
+      if not re.match(filename_regex, filename):
+        continue
+      full_path = os.path.join(path, filename)
+      if os.access(full_path, os.X_OK):
+        yield full_path
+
+
+def find_clang_format(min_major_version):
+  '''Find the first clang-format with a minimum major version.'''
+  clang_formatters = find_executable("clang-format(-\d+)?$")
+  for clang_formatter in clang_formatters:
+    version = subprocess.check_output([clang_formatter, "--version"])
+    major_version = re.search("\d+", str(version))
+    if int(major_version.group()) >= min_major_version:
+      return clang_formatter
+
+
+def get_changed_files():
+  '''Find all changed files in the index.
+
+  Yield each file in the format (octal_mode, filename).
+
+  '''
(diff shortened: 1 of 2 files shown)


eyal0:
In my testing it works okay but there are a lot of corner cases so I might have missed some.  Anyway, I think that it's a decent first start.  We might start with this and, over time, correct any errors.  I don't think that there's any big risk in trying this out even if it has some bugs.  At worst, we'll catch the errors in CI.

@henrygab @ooxi

ooxi:
Thank you very much @eyal0!

henrygab:
It appears there's now a PR / CI check that ensures the formatting.  Awesome!

Can you help me understand what this 'pre-commit' hook does?

Specifically, can you confirm that `git` will not reformat automatically?
e.g., the user has to manually launch the reformatting?

I ask because having the format occur automatically (e.g., at `git add`, `git push`, etc.) is ... really disruptive.  I often have non-conformant formatting while working on an issue; it's just much easier for me to think through.

Thus, if my staged changes get automatically changed as part of running `git add` or similar, I will be extremely frustrated.   Hopefully, I've misunderstood?

eyal0:
Oh boy, you are not going to be happy about this!

Well, first of all, it is not possible to to install a pre-commit hook for someone else.  Anyone that clones this repo will *not* get the pre-commit hook by default.  To install the pre-commit hook, you can run the script `install-hooks` and that will install it.

After it's installed, it *will* reformat automatically for you.  But we can change that!

How would you like it to work.  Do you want the pre-commit hook to just cause your attempt to commit to fail when you make a commit that has bad formatting?  Or something else?  Would you like it to print something out, like, the errors that clang finds?

The code that I wrote is pretty flexible so whatever idea you have, I'm sure that I could implement it.

henrygab:
Aha!  I'm glad to hear it cannot install a pre-commit hook silently.  _(that would be worrying!)_  The CI check will remind of the problem for me, if I forget to check ahead of time.

So long as it's not required on my dev box, I'm OK.  :)   Thanks for helping me understand (yet another) feature of `git`! 👍

PR #204, opened 2023-08-24
Move pre-commit hook into utils directory

ooxi:
In order to keep the root directory clean, I propose to move the pre-commit hook into the utils directory. Moreover I explicitly stated the required Python version and added the clang-format dependencies to isolated build environments.

@eyal0 are you ok with this change?

See PR #203


ooxi:
I want ahead and merged this, feel free to undo

PR #206, opened 2023-08-26
Update tests, .gitignore

henrygab:
`test/run_tests.sh` was updated:
* adding an option to specify a custom test list (useful for targeted testing)
* updating the displayed help text

`test/tests.list` was updated:
* Clarify the built-in documentation at the top of the file.
* Replace `+` with underscore for two test identifiers
* Replace `+` with underscore in name of generated file

Generally, '+' is frowned upon within filenames on Windows,
as it is considered a special character by `cmd.exe` (used to
concatenate two files together).  This means any use in `cmd.exe`
requires special care ... and most folks don't even know how to
escape a single character in `cmd.exe` (`^`).

As many shells have unique constraints on filenames, special
characters, etc., enforcing a limited character set here is just
the right thing to do....


.gitignore
```diff
@@ -65,3 +65,10 @@ test/mismatch/
 test/outputs/
 win32/BUGS.txt
 win32/COPYING.txt
+
+# Visual Studio code files
+.vscode/
+
+# files generated by running test/run_tests.sh
+test/inputs/test-image-rotation-1-rotated.gbx
```
eyal0, line 73: Thanks for this, it had always been so annoying. It would be nicer if those temporary files were built into an outputs directory and e could ignore that whole directory. But this is fine, too.

If one day we have lots of these then we'll rethink it.

eyal0:
SKIPPED: No reference file ./golden/test-merge-a_b.png

That's the error...

henrygab:
> 

:)  Just noticed it myself, and was preparing a fix.  (now pushed)   :)

ooxi:
Thank you very much @henrygab! If a commit in PR fails I would prefer to force push a fixed commit instead of adding a fixing commit to the PR :-)

henrygab:
> Thank you very much @henrygab! If a commit in PR fails I would prefer to force push a fixed commit instead of adding a fixing commit to the PR :-)

Ok.  Thanks for letting me know this ... will do so in future.

eyal0:
> Thank you very much @henrygab! If a commit in PR fails I would prefer to force push a fixed commit instead of adding a fixing commit to the PR :-)

Yeah, that's good practice. It's nice if every commit that eventually hits mains passes all unit tests. That way, if we use git bisect in the future, we won't accidentally fail on an unrelated issue.

PR #210, opened 2023-09-04
Fix multiple issues

henrygab:
Highlights are as follows:

1. Fix MIN/MAX macros ... they only worked from C++, failed in C11.  See https://github.com/gerbv/gerbv/commit/45bc71acd855787eebb5aa9403baa8be42260266
1. `clang-format` CI check was not applying the checked-in file style.  See https://github.com/gerbv/gerbv/commit/0f74a9ce107174613392dfb7fc04c63cb665fd4f.
1. Convert unsage `dprintf()` to safer `DPRINTF(())`.  See https://github.com/gerbv/gerbv/commit/a831563a667e72d4e87aa7296bbfec0f6b103f5d
1. Fix issue #195.  See https://github.com/gerbv/gerbv/commit/c261beb39de269f311b31da073c50b8f326fed75
1. Add checks for invariants.  See https://github.com/gerbv/gerbv/commit/d3c32dfdbddd1111f10491891e5cf108f36df009 and https://github.com/gerbv/gerbv/commit/f5d9c723f9123002999a20c99c9f1eef7e6a107b

All commits build and pass tests.  See https://github.com/henrygab/gerbv/pull/1 for automated build / test results for each commit.

Fixes #195.
Fixes #211.
Fixes #212.
Fixes #213.
Fixes #214.
Fixes #215.
Fixes #216.
Fixes #217.
Fixes #218.
Fixes #219.
Fixes #220.
Fixes #221.

PR #223, opened 2023-10-11
MIN/MAX only valid in C++

eyal0:
* C11 does not support references.
* C11 does not support the `typeof` operator (although GCC allows it as an extension)


eyal0:
Commit by commit, chipping away at #210

PR #240, opened 2024-07-23
Fix SVG export scale for Cairo 1.17.6+

### LINKED ISSUE #231: Gerbv scales down when I export to SVG
I have this example gerber file with two "circles". The distance between them is 2.54mm. If I export to PDF and then open in Inkscape the distance is still 2.54mm. But if I export to SVG the distance is scaled down to 1.91mm. Everything is scaled down and you can't realize it until you measure it or print it. The same happens with the Print option.

This happens in Arch Linux with the latest version (from AUR) 2.10.0 but also with 2.9.5. I had to use Windows 7 with this version gerbv-win-static_20160713 to properly export my files.
[samples.zip](https://github.com/gerbv/gerbv/files/13619507/samples.zip)

The bug is present in version gerbv-2.7.0. This is the oldest version available to download from archive.archlinux.org so I can't test earlier versions.

PingPongun:
Fixes #231

### DIFF
--- a/src/draw.c
+++ b/src/draw.c
@@ -775,6 +775,15 @@ draw_image_to_cairo_target(
     gboolean doVectorExportFix;
     double   bg_r, bg_g, bg_b; /* Background color */
 
+#if CAIRO_VERSION >= CAIRO_VERSION_ENCODE(1, 16, 0)
+    // Fix for cairo 1.17.6 and above which sets to surface unit back to PT (default: UNIT_USER)
+
+    cairo_surface_t* cSurface = cairo_get_target(cairoTarget);
+    if (cairo_surface_get_type(cSurface) == CAIRO_SURFACE_TYPE_SVG) {
+        cairo_svg_surface_set_document_unit(cSurface, CAIRO_SVG_UNIT_PT);
+    }
+#endif
+
     doVectorExportFix = draw_do_vector_export_fix(cairoTarget, &bg_r, &bg_g, &bg_b);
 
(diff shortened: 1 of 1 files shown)


spe-ciellt:
Thank you for finding the problem and submitting a fix. I have one question before merging this. Is the #if necessary?  If you have that line using an older version of libcairo, will it break anything? I would like to have that patch, but without the #if/#endf (if possible). It feels better to always make sure we use the correct units. Maybe a comment on the Cairo versioning thing?

PingPongun:
The function `cairo_svg_surface_set_document_unit`
Has been added in 1.16 so without `if`s it will break for Cairo bellow 1.16

spe-ciellt:
OK, great, then I know. It looks a bit like the formatting is a bit off. The #if/#endif should be on the column and the code added should have spaces and not tabs (I guess you are using). If you look at the diff in Github you see the difference.
We have a .clang-format, so you can use that to check the formatting.

PingPongun:
I believe formatting should now be correct

PingPongun:
Corrected (and this time (previously I was to lazy) I've also built it locally and tested- it fixes issue at least for me)

PingPongun:
I've updated code so that it will also work when using `File>>Print` (previous one worked with cli export and `File>>Export>>SVG`, as mentioned in https://github.com/gerbv/gerbv/issues/231#issuecomment-2267650758)

spe-ciellt:
@PingPongun  I merged this now. Thank you for your patch and rework!

PR #268, opened 2025-12-23
Test Github Actions PR, not to be merged

spe-ciellt:
Add step in CI to check if we have a PAT


spe-ciellt:
This was a test of the pipeline. I kill it off at the moment so it doesn't pollute the PR space.

PR #274, opened 2026-02-13
feat: Add "Fast, with OR", a new, faster rendering mode.

### LINKED ISSUE #273: tip: a fast (X)OR renderer with slightly nicer output
This is not a bug report, but rather a tip to improve the XOR renderer.
The Normal and High quality renderers are extremely slow on my new 14th gen Intel Core and Ubuntu 22.04. I managed to improve the visual appeal of the (X)OR renderer so that the output looks passable - at least to my liking - while maintaining the speed of the fast renderer.

Here is the git diff of the change:
<pre>
diff --git a/src/gerbv.c b/src/gerbv.c
index e7c1f3f..5c66283 100644
--- a/src/gerbv.c
+++ b/src/gerbv.c
@@ -808,6 +808,7 @@ gerbv_render_to_pixmap_using_gdk (gerbv_project_t *gerbvProject, GdkPixmap *pixm
        */
        for(i = gerbvProject->last_loaded; i >= 0; i--) {
                if (gerbvProject->file[i] && gerbvProject->file[i]->isVisible) {
+                       int cR, cG,cB;
                        /*
                        * Fill up image with all the foreground color. Excess pixels
                        * will be removed by clipmask.
@@ -825,7 +826,17 @@ gerbv_render_to_pixmap_using_gdk (gerbv_project_t *gerbvProject, GdkPixmap *pixm
                                gdk_gc_set_function(gc, GDK_COPY);
                        }
                        else if (renderInfo->renderType == GERBV_RENDER_TYPE_GDK_XOR) {
-                               gdk_gc_set_function(gc, GDK_XOR);
+                               cR = (gerbvProject->file[i]->color.pixel >> 16) & 0xFF;
+                               cG = (gerbvProject->file[i]->color.pixel >> 8) & 0xFF;
+                               cB = (gerbvProject->file[i]->color.pixel) & 0xFF;
+                               if ( 
+                                       ((cR == cB) && (cG <= cR)) || // violet or gray -> use copy function
+                                       ((cR == cG) && (cB <= cR))     // yellow -> use copy function
+                               ) {
+                                       gdk_gc_set_function(gc, GDK_COPY);
+                               } else {
+                                       gdk_gc_set_function(gc, GDK_OR);
+                               }
                        }
                        /*
                        * Translation is to get it inside the allocated pixmap,

</pre>

The trick is to combine COPY and OR drawing functions on individual layers based on their color. The layers defined with yellow or violet or gray color are treated with COPY function the rest of the layers is OR ed. Ideally, the OR or COPY could be defined the same way the transparency is defined in the color picker, but for the sake of simplicity the careful selection of colour does the job.

Here is an example of the output - you can still see the 'semitransparent' tracks while the pads and solder mask is opaque.

Image

Hope it helps.

eyal0:
This fixes #273.

### DIFF
--- a/src/callbacks.c
+++ b/src/callbacks.c
@@ -1063,7 +1063,7 @@ callbacks_toggle_layer_visibility_activate (GtkMenuItem *menuitem, gpointer user
 
 	callbacks_update_layer_tree ();
 
-	if (screenRenderInfo.renderType <= GERBV_RENDER_TYPE_GDK_XOR) {
+	if (uses_gdk(screenRenderInfo.renderType)) {
 		render_refresh_rendered_image_on_screen ();
 	} else {
 		render_recreate_composite_surface ();
@@ -2182,7 +2182,7 @@ callbacks_layer_tree_visibility_toggled (gint index)
 		!mainProject->file[index]->isVisible;
 
 	callbacks_update_layer_tree ();
-	if (screenRenderInfo.renderType <= GERBV_RENDER_TYPE_GDK_XOR) {
+	if (uses_gdk(screenRenderInfo.renderType)) {
 		render_refresh_rendered_image_on_screen ();
 	} else {
 		render_recreate_composite_surface ();
@@ -2434,7 +2434,7 @@ callbacks_remove_layer_button_clicked (GtkButton *button, gpointer user_data)
 		index = MAX(0, index - 1);
 		callbacks_select_layer_row (index);
 
-		if (screenRenderInfo.renderType <= GERBV_RENDER_TYPE_GDK_XOR) {
+		if (uses_gdk(screenRenderInfo.renderType)) {
 			render_refresh_rendered_image_on_screen ();
 		} else {
 			render_recreate_composite_surface ();
@@ -2464,7 +2464,7 @@ callbacks_move_layer_down_button_clicked  (GtkButton *button, gpointer   user_da
 		callbacks_update_layer_tree ();
 		callbacks_select_layer_row (index + 1);
 
-		if (screenRenderInfo.renderType <= GERBV_RENDER_TYPE_GDK_XOR) {
+		if (uses_gdk(screenRenderInfo.renderType)) {
 			render_refresh_rendered_image_on_screen ();
 		}
 		else {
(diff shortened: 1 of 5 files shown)


eyal0:
Please wait to merge this until https://github.com/gerbv/gerbv/issues/273#issuecomment-3898475559 is satisfied.  I would like to see the speed improvement first.

spe-ciellt:
Not sure if this is just a proof of concept or actual implementation. Please note that if/when `gerbv` is ported to GTK-3, all GDK drawing modes will disappear since GTK3 does not support GDK.

But here are some comments anyhow.

---

### Hard blocker — enum insertion order

`GERBV_RENDER_TYPE_GDK_OR` is inserted at position 2, shifting the Cairo types:

| Enum value | Before | After |
|---|---|---|
| 0 | GDK | GDK |
| 1 | GDK_XOR | GDK_XOR |
| 2 | CAIRO_NORMAL | **GDK_OR** |
| 3 | CAIRO_HIGH_QUALITY | CAIRO_NORMAL |
| 4 | MAX | CAIRO_HIGH_QUALITY |
| 5 | — | MAX |

Any user who had Cairo Normal (2) or Cairo High Quality (3) saved in GSettings or a `.gvp` project file will silently load the wrong mode after upgrading. **The fix is to append `GDK_OR` before `MAX` rather than inserting it mid-sequence**, and update `uses_gdk()` to use explicit comparisons instead of `t <= GERBV_RENDER_TYPE_GDK_OR`.

---

### Other issues

**Stale comment in `gerbv.c`:**
```c
/* renderTypes 0 and 1 use GDK rendering, so we shouldn't have made it
   this far */
```
After the PR, types 0, 1, and 2 are GDK. Comment must be updated.

**`uses_gdk()` ordering contract:** The helper relies on GDK modes being numerically lowest:
```c
return t <= GERBV_RENDER_TYPE_GDK_OR;
```
This is a silent contract that breaks if modes are ever reordered. A switch or explicit OR of the three GDK constants would be more robust.

**Color threshold heuristic is fragile:** The logic picks `GDK_COPY` vs `GDK_OR` based on whether all RGB channels are non-zero. A nearly-black color like `{1, 1, 1}` (out of 65535) would get `GDK_COPY` even though it's visually indistinguishable from black. The rationale deserves a code comment, and the known limitation with non-black backgrounds (OR compositing produces garbage unless the background is black) should be documented.

**Commit message typo:** "Create a `uses_gdb` helper function" — should be `uses_gdk`.

eyal0:
Good comments.  For now, I've yet to see evidence that this PR is necessary.  I wrote this PR so that we can test the idea in #273.  I'm not yet convinced that this is necessary because XOR seems at least as good as this.

If we get solid evidence that it's worthwhile to add this then I'll address the comments and we can try to merge.  If we don't find a solid need for this then we can abandon this.

PR #275, opened 2026-02-21
fix: set omit_zeros unconditionally when LZ/TZ is declared in Excellon header

### LINKED ISSUE #174: v2.9.6 Trailing zero imports wrong
In v2.9.6 (Ubuntu Lunar 23.04) I have:
```
M48
;Layer_Color=9474304
;FILE_FORMAT=2:5
INCH,LZ
;TYPE=PLATED
T1F00S00C0.00800
T2F00S00C0.01200
%
T01
X0060743Y0023895
X006281Y0025273
X0062909Y0022615
...
```
The (Altium generated) drill file is auto detected as None/Inch/5. Obviously 1st and 3rd drill are shown correctly, 2nd drill is placed near the origin (x = 0.06281).

Manually setting Trailing/Inch/2 all drills disappear entirely.

This is a regression as in 2.7.x and 2.8.x this worked correctly.

rampageservices:
## Summary

Fixes #174. When an Excellon drill file contains both `FILE_FORMAT` and `INCH,LZ` (or `METRIC,LZ`, `INCH,TZ`, etc.), the `FILE_FORMAT` handler sets `autod = 0`, which causes the subsequent LZ/TZ parsing to skip setting `omit_zeros`. This results in incorrect coordinate interpretation — trailing zeros are not suppressed when they should be.

The fix moves the `omit_zeros` assignment outside the `if (state->autod)` gate in both `drill_parse_header_is_metric()` and `drill_parse_header_is_inch()`. The `autod` gate is preserved for `decimals` and `number_format` — only `omit_zeros` is decoupled.

This regression was introduced in 944a4b3 (PR #156).

## What changed

- `src/drill.c`: 4 lines moved — `omit_zeros` assignment for cases 'L' and 'T' in both `drill_parse_header_is_metric()` and `drill_parse_header_is_inch()` now executes unconditionally
- `test/golden/LimeSDR-QPCIe_1v2-RoundHoles.png`: Regenerated — previous golden was produced with the buggy `autod=0` path
- `test/inputs/Altium_file_format_inch_lz.drl`: New test input — `FILE_FORMAT=2:5` before `INCH,LZ` with trailing-zero-suppressed coordinates
- `test/tests.list`: Added `Altium_file_format_inch_lz` test entry

## Regression analysis

| Scenario | autod | Before fix | After fix |
|----------|-------|-----------|-----------|
| `INCH,TZ` alone (no FILE_FORMAT) | 1 | omit_zeros set | omit_zeros set — **same** |
| `FILE_FORMAT` then `INCH,TZ` | 0 | omit_zeros **skipped** (bug) | omit_zeros set — **fixed** |
| `FILE_FORMAT` then `METRIC,LZ` | 0 | omit_zeros **skipped** (bug) | omit_zeros set — **fixed** |
| `INCH,TZ` then `FILE_FORMAT` | 1→0 | omit_zeros set (autod still 1 during TZ parse) | omit_zeros set — **same** |
| No LZ/TZ at all | any | fallback at header-end | **unchanged** |

The new test `Altium_file_format_inch_lz` fails on develop with MAE=13726 and passes with this fix.

## Test plan

- [x] All existing drill tests pass (except pre-existing `test-drill-trailing-zero-suppression` failure unrelated to this change)
- [x] New `Altium_file_format_inch_lz` test passes
- [x] Verified new test fails on develop without the fix
- [x] `Altium_file_format_inch` and `Altium_inch_file_format` unchanged (still pass)


spe-ciellt:
Hello and thank you for the patch. I will merge this patch since it seems to solve a real world problem and does not affect the current ones.

PR #279, opened 2026-02-24
feat: add optional Inkscape-layer SVG export

5shekel:
note: this is generated code, prompt as follows
> add option in svg export to output all gerber layers as svg layers, in inkscape style.

given the quality of this project, and lack of deep understanding from my side i cannot say this is valid PR. 
it does work on my tests and is a feature i was craving for ages.  
might be of interest to others. 

image

---
Add a new SVG export option that preserves visible Gerber layers as Inkscape layers so exported artwork remains editable by layer. Wire the option into both CLI and GUI export flows and document usage in help/man text.

### DIFF
--- a/src/export-image.c
+++ b/src/export-image.c
@@ -42,6 +42,140 @@
 #include <cairo-pdf.h>
 #include <cairo-ps.h>
 #include <cairo-svg.h>
+#include <glib/gstdio.h>
+
+static gchar *
+exportimage_extract_svg_inner_content (const gchar *svgText) {
+	const gchar *svgOpen;
+	const gchar *svgOpenEnd;
+	const gchar *svgClose;
+
+	if (svgText == NULL)
+		return NULL;
+
+	svgOpen = strstr (svgText, "<svg");
+	if (svgOpen == NULL)
+		return NULL;
+
+	svgOpenEnd = strchr (svgOpen, '>');
+	if (svgOpenEnd == NULL)
+		return NULL;
+
+	svgClose = g_strrstr (svgOpenEnd, "</svg>");
+	if (svgClose == NULL || svgClose <= svgOpenEnd)
+		return NULL;
+
+	return g_strndup (svgOpenEnd + 1, svgClose - (svgOpenEnd + 1));
+}
+
+static void
+exportimage_append_svg_background (GString *svgOut, gerbv_project_t *gerbvProject) {
+	GdkColor *bg = &gerbvProject->background;
+
+	/* Keep legacy vector-export behavior: skip solid white/black backgrounds. */
+	if ((bg->red == 0xffff && bg->green == 0xffff && bg->blue == 0xffff)
+	 || (bg->red == 0x0000 && bg->green == 0x0000 && bg->blue == 0x0000))
+		return;
+
+	g_string_append_printf (svgOut,
+		"  <rect x=\"0\" y=\"0\" width=\"100%%\" height=\"100%%\" fill=\"rgb(%u,%u,%u)\" />\n",
+		bg->red / 257, bg->green / 257, bg->blue / 257);
+}
+
+static void
+exportimage_render_layer_to_svg_file (gerbv_fileinfo_t *fileInfo,
+		gerbv_render_info_t *renderInfo, const gchar *filename) {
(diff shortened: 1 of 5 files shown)


spe-ciellt:
Hello @5shekel 

Thank you for you pull request. Seems like a useful thing. 

Tested the branch — builds clean, no regressions (89/100 pass, same 11 failures as `develop` baseline). That is because Cairo versioning thing, but it builds and is accepted by the regression tests in the pipeline. The feature works correctly; layered SVG output is well-formed and opens with proper Inkscape layers.

A few things worth addressing:

**1. Nested `<svg>` inside layer groups**
The output wraps each layer in a child `<svg>` inside the `<g>`, which is non-standard for Inkscape layers. Standard practice is to put drawing primitives directly in the `<g>`. It works in Inkscape but may behave oddly in other SVG editors. Not a blocker for a first version.

**2. Partial output on error**
If `exportimage_render_svg_layers_from_project()` fails mid-loop, it still writes the partial SVG to disk. Should either skip writing or `g_unlink()` the output file on error.

**3. Silent background skip**
`exportimage_append_svg_background()` silently skips pure white and pure black — worth a comment explaining why, since it'll surprise anyone who deliberately set a white background.

**Final note**
When you fix this, could you merge all commits into one and then force-push it to this branch? (`git push --force-with-lease`)

src/export-image.c
```diff
@@ -42,6 +42,140 @@
...
+	if (svgClose == NULL || svgClose <= svgOpenEnd)
+		return NULL;
+
+	return g_strndup (svgOpenEnd + 1, svgClose - (svgOpenEnd + 1));
+}
+
+static void
+exportimage_append_svg_background (GString *svgOut, gerbv_project_t *gerbvProject) {
+	GdkColor *bg = &gerbvProject->background;
+
```
spe-ciellt, line 78: Please elaborate on why this colors are changed. And also please use {} even if it is an one-liner on the `if`, `for`, `while`, etc.

src/export-image.c
```diff
@@ -42,6 +42,140 @@
...
+static gchar *
+exportimage_extract_svg_inner_content (const gchar *svgText) {
+	const gchar *svgOpen;
+	const gchar *svgOpenEnd;
+	const gchar *svgClose;
+
+	if (svgText == NULL)
+		return NULL;
+
+	svgOpen = strstr (svgText, "<svg");
```
spe-ciellt, line 57: See comment on wrapping inner block `<svg>...</svg>`.

spe-ciellt (commented):
I tried to do a review. This is really the first time using the github interface, so please bear with me. Else I just let the comment I wrote speak for itself.

spe-ciellt:
Re-checked after the two follow-up commits (`8da9640`, `aa22e7d`). All three issues from the earlier review have been addressed:

**1. Nested `<svg>` inside layer groups — Fixed**
`exportimage_extract_svg_inner_content()` now extracts only what's *between* the Cairo `<svg>` opening tag and `</svg>`, so each Inkscape layer `<g>` contains drawing primitives directly. No more nested `<svg>`.

**2. Partial output file on error — Fixed**
The function builds the entire SVG in a `GString` in memory and only writes to disk at the end, guarded by `!hadError`. A failed layer mid-loop leaves no partial file on disk.

**3. Undocumented background skip — Fixed**
The comment in `exportimage_append_svg_background()` now explains that white/black are intentionally skipped to stay consistent with the non-layered SVG export path.

One minor item still open: `exportimage_extract_svg_inner_content()` uses `strstr`/`strchr`/`g_strrstr` to slice up Cairo's SVG output. A short comment noting that this relies on Cairo's stable output format would help future maintainers. Not a blocker.

Tested the updated branch against `example/hellboard/` (3 layers):

- Builds clean
- Output is valid XML; each `<g inkscape:groupmode="layer">` contains `<path>` elements directly — no nested <svg>`
- Layer z-order matches gerbv's normal rendering order (`last_loaded` → 0)
- Cairo does not emit `<defs>`/`clip-path` for Gerber content, so the string-extraction approach loses nothing in practice (though worth a comment if that ever changes)
- Pixel-for-pixel identical to the flat SVG export (ImageMagick `compare -metric AE` = 0)

Otherwise looks good — backward-compatible, feature works correctly.

PR #282, opened 2026-02-26
fix: silently ignore Gerber X2 attribute commands (TF, TA, TO, TD)

### LINKED ISSUE #249: Unknown RS-274X extension found
Recent versions of KiCad (>=8?) have started creating Gerber files that have RS-274X extensions %TA% %TD% %TF% %TO%.
When loaded by gerbv, the rendering is still perfectly fine, but it fills the console and the Messages pane with complaints like
```
Unknown RS-274X extension found %TF% at line 3 in file "/redacted/test1.gbr"
```

I found a pretty small example that I attach here: [zest-Edge_Cuts.txt](https://github.com/user-attachments/files/19542537/zest-Edge_Cuts.txt)

The only mention of this issue I could find online was in an out-of-the-way place:
X2 extensions should be ignored, by react66, Aug 23, 2021
https://github.com/alyoshin/gerbv-dxf-builded/issues/1
Note the date: over four years ago! Maybe this issue will be better at reaching the right people.
In that issue, react66 included the helpful comment:
> Gerber X2 has adds four new commands (TF, TA, TO and TD) to attach attributes to a Gerber file. All pre-existing commands remain unchanged.

I can confirm the effect on the Debian-shipped gerbv versions 2.9.6 and 2.10.0.
I haven't yet built gerbv from github source to find out if it silently got improved.

rampageservices:
## Summary

KiCad >= 8 and other modern EDA tools generate Gerber X2 files containing `%TF%`, `%TA%`, `%TO%`, and `%TD%` attribute commands. These are standard Gerber X2 metadata extensions (defined by Ucamco in 2014) that don't affect rendering — gerbv already renders these files correctly — but `parse_rs274x()` doesn't recognize them, causing error-level spam:

```
Unknown RS-274X extension found %TF% at line 3 in file "..."
```

This PR adds four `case` statements to match these commands and log them at `GERBV_MESSAGE_NOTE` level instead of `GERBV_MESSAGE_ERROR`. The existing cleanup code (lines 1853-1858) already reads until the trailing `*` character, so the attribute data is consumed automatically — only the `case`/`break` is needed.

### Gerber X2 Attribute Commands

| Command | Name | Purpose |
|---------|------|---------|
| `%TF.<name>,<values>*%` | File Attribute | Metadata about the file (software version, creation date, etc.) |
| `%TA.<name>,<values>*%` | Aperture Attribute | Metadata attached to aperture definitions |
| `%TO.<name>,<values>*%` | Object Attribute | Metadata attached to graphical objects |
| `%TD[.<name>]*%` | Delete Attribute | Removes a previously set attribute |

### Changes

- `src/gerber.c`: Add `case A2I('T','F')` / `A2I('T','A')` / `A2I('T','O')` / `A2I('T','D')` to the switch in `parse_rs274x()`, logging at NOTE level and breaking
- `test/inputs/test-gerber-x2-attributes.gbx`: Minimal Gerber X2 test file with all four attribute types
- `test/golden/test-gerber-x2-attributes.png`: Golden reference for the test
- `test/tests.list`: Register the new test

Fixes #249

## Test plan

- [x] Build gerbv from modified source (`cmake .. && make`)
- [x] Load test Gerber X2 file — no error messages, only NOTE-level log
- [x] Run test suite (`./run_tests.sh`) — new test passes, no regressions


spe-ciellt:
Thank you for this patch. You've been busy it seems like. ;) I will work through your list of PRs.  If they  are good enough I apply them else I will leave comments.

spe-ciellt:
Tested on the branch. Build is clean, all four attribute types (`TF`, `TA`, `TO`, `TD`) are handled correctly.     
                                                                                             
**Code:** The fix slots cleanly into the existing switch pattern. The unconditional cleanup code after the switch (`gerb_ungetc` + read-until-`*` loop) consumes the attribute content automatically, so only the `case`/`break` is needed — no extra parsing. The `%TD*%` edge case (no attribute name, `*` immediately follows) also works correctly. Message format and `gerbv_escape_char` usage are consistent with the rest of the function.
                                                                                                                      
**Tests:** New test `test-gerber-x2-attributes` passes. 91/101 pass overall; the 10 failures are the same Cairo-version mismatches as the `develop` baseline — no regressions.
                               
The NOTE vs ERROR distinction is only visible in the GUI Messages pane, not CLI output, so it can't be exercised by the automated suite — but the rendered output test is the right check here.

PR #285, opened 2026-02-26
Fix aperture macro 0X parsed as hex float instead of multiplication

### LINKED ISSUE #253: Aperture Macro Rotation Bug
Paste layer, rendered in Cuprum:

Image

Tracespace.io:

Image

KiCad's gerber viewer:

Image

And finally, gerbv (obviously broken):

Image

The lines causing these problems are all something along the lines of:

20,1,$1+$1,$2,$3,$4,$5,0X25.4+90*

Yes, It's weird and not actually needed, but CAD programs are notoriously stupid when it comes to the sort of output they sometimes generate, especially for things like gerbers.

In this case, the rotation (the final argument) should be 90 degrees, as 0x25.4+90 = 90 degrees. However, I manually measured the angle of the odd pad rotation gerbv displays and it is 36 degrees off from the correct 0 degree angle.

25.4*90 mod 90 = 36. The only way rotations would end up like that is if gerbv thinks 0X25.4+90 = 2286, which it isn't, it's 0. Given that  gerbv is the only library/program I can find with this problem, I think it's safe to say this is a bug specific to gerbv and not something that can be blamed on the gerber being 'unusual'.

rampageservices:
## Summary

- Fixes `gerb_fgetdouble()` misparsing expressions like `0X25.4+90` as hexadecimal floating-point literals (C99 `strtod` interprets `0X25.4` as 37.25) instead of letting the aperture macro parser handle `X` as the multiplication operator
- Adds an early check for `0x`/`0X` prefix that returns `0.0` and advances past only the `0`, so the expression correctly evaluates as `0 × 25.4 + 90 = 90`
- This is safe for all callers since Gerber format never uses hexadecimal numbers

## Test plan

- [x] New test file `test-amacro-hex-expression.gbx` exercises the `0X25.4+90` expression in a LINE21 rotation parameter
- [x] Golden PNG `test-amacro-hex-expression.png` included — shows rectangle at 90° rotation (not ~127°)
- [x] RS-274X round-trip export confirms rotation = 90.000000 (not 127.250000)
- [x] Full test suite: 90 passed (up from 89), 0 new failures

Fixes #253


spe-ciellt:
Thank you for this patch. If this is good enough I apply it else I will leave comments.

spe-ciellt:
Looks simple enough. There is the reading more data than maybe available, but  I guess it is done in more places in gerbv. An overhaul on how data is read might be in its place.
Thank you for spotting this and I will merge it.

PR #292, opened 2026-02-27
Wire NOTE messages into CLI output with severity prefixes

### LINKED ISSUE #291: CLI output should differentiate message severity levels
## Problem

In CLI (non-GUI) mode, gerbv does not differentiate between message severity levels in its output. All messages from the GLib log handler are printed to stderr with the same format, making it impossible to distinguish errors from warnings or notes.

As noted during review of #282:

> The NOTE vs ERROR distinction is only visible in the GUI Messages pane, not CLI output, so it can't be exercised by the automated suite.

The type system already exists — `gerbv_message_type_t` defines four levels (FATAL, ERROR, WARNING, NOTE) and `error_type_string()` in `callbacks.c` maps them to display strings. But this infrastructure is only used in the GUI.

### Message type enum (`gerbv.h:138-142`)

```c
typedef enum {
    GERBV_MESSAGE_FATAL,   // processing cannot continue
    GERBV_MESSAGE_ERROR,   // something went wrong, but processing can continue
    GERBV_MESSAGE_WARNING, // something may provide wrong output
    GERBV_MESSAGE_NOTE     // irregularity, no intervention needed
} gerbv_message_type_t;
```

### GLib log macros (`gerbv.h:120-123`)

```c
#define GERB_FATAL_ERROR(...)     g_log(NULL, G_LOG_LEVEL_ERROR, __VA_ARGS__)
#define GERB_COMPILE_ERROR(...)   g_log(NULL, G_LOG_LEVEL_CRITICAL, __VA_ARGS__)
#define GERB_COMPILE_WARNING(...) g_log(NULL, G_LOG_LEVEL_WARNING, __VA_ARGS__)
#define GERB_MESSAGE(...)         g_log(NULL, G_LOG_LEVEL_MESSAGE, __VA_ARGS__)
```

There is **no macro for NOTE-level** messages. Notes are stored in the error list via `gerbv_stats_add_error()` but are not output to stderr in CLI mode.

### CLI message handler (`callbacks.c:375-386`)

The temporary handler (`callbacks_temporary_handle_log_messages`) stores messages and delegates to `g_log_default_handler()`, which outputs to stderr without severity prefixes.

### GUI handler (`callbacks.c:3994-4081`)

The GUI handler color-codes messages by level (red for errors, darkblue for messages, darkgreen for info, etc.) and routes them to the Messages pane. This distinction is lost in CLI mode.

### Unused `--log` flag

The `-l, --log=<logfile>` CLI option is parsed (`main.c:787-794`) but the `logToFileOption` and `logToFileFilename` variables are **never used** — file logging is unimplemented.

### 1. Add severity prefixes to CLI output

Replace or wrap the temporary log handler with a CLI-specific handler that prefixes messages:

```
error: Unknown RS-274X extension found %XX% at line 3 in file "..."
note: Gerber X2 file attribute %TF.GenerationSoftware% (ignored)
warning: Trailing garbage found after numeric data
```

### 2. Add a `--verbosity` flag

```
--verbosity=<level>   Set minimum message level: fatal, error, warning, note (default: note)
--quiet               Suppress note-level messages (equivalent to --verbosity=warning)
```

### 3. Implement the `--log` flag

The flag is already parsed but unused. Wire it up to write formatted messages to the specified file.

### 4. Add a `GERB_NOTE()` macro

```c
#define GERB_NOTE(...) g_log(NULL, G_LOG_LEVEL_INFO, __VA_ARGS__)
```

This gives NOTE-level messages a path to stderr output, consistent with the other macros.

## Impact

- Makes CLI output machine-parseable (grep for `error:` vs `note:`)
- Enables test suites to verify message levels (currently impossible per #282 review)
- The `--quiet` flag would suppress the Gerber X2 attribute notes that originally prompted #249
- Completing `--log` finishes an existing but broken feature

## Related

- #282 — Gerber X2 attributes (review noted CLI limitation)
- #249 — Unknown RS-274X extension found (X2 attribute error spam)

rampageservices:
## Summary

- Add `GERB_NOTE` macro (`G_LOG_LEVEL_INFO`) and emit NOTE messages through `g_log` in `gerb_stats.c`
- Replace `g_log_default_handler` passthrough with a custom formatter that prints clean `fatal:`/`error:`/`warning:`/`note:` prefixes to stderr
- Add `--quiet`/`-q` flag to suppress note-level messages
- Wire up the existing `--log`/`-l` flag to actually write prefixed messages to a file

## Context

PR #282 (Gerber X2 attributes) added `GERBV_MESSAGE_NOTE` messages for X2 attribute commands. The `NOTE` case in `gerbv_stats_add_error()` had a `break` with no `g_log()` call — messages were stored in the error list but never emitted to stderr. Messages that did reach stderr used GLib's default format (`** (process): WARNING **: text`) with no clean prefix. The `--log` flag was parsed but never connected.

## Changes

**`src/gerbv.h`** — `GERB_NOTE(...)` macro mapping to `G_LOG_LEVEL_INFO`

**`src/gerb_stats.c`** — Wire empty `GERBV_MESSAGE_NOTE` case to `GERB_NOTE()`

**`src/main.c`**:
- Custom `callbacks_temporary_handle_log_messages` that formats with severity prefixes, respects `--quiet`, writes to log file, and preserves `log_array_tmp` for GUI replay
- `--quiet`/`-q` option (longopts, opt_options, switch case, help text)
- `--log`/`-l` file open after option parsing, `fclose` on both CLI exit and GUI return paths

## Test plan

- [x] Build: `cd build && cmake .. && make` — clean, no warnings
- [x] X2 file without `--quiet`: shows `note: Ignoring Gerber X2 attribute...`
- [x] X2 file with `--quiet`: note suppressed, no output
- [x] Bad file: `error:` and `warning:` prefixes shown correctly
- [x] `--log /tmp/gerbv.log`: log file written with prefixed messages
- [x] Test suite: 94 passed, 10 pre-existing rendering mismatches (unrelated)

Fixes #291

### DIFF
--- a/man/gerbv.1.in
+++ b/man/gerbv.1.in
@@ -35,7 +35,11 @@ with Cairo) it can be specified as an #RRGGBBAA code. Use multiple
 \-f flags to set the color for multiple layers.
 .TP
 .BI -l\ <filename>|--log=<filename>
-All error messages etc are stored in a file with filename \fI<filename>\fP.
+Write all messages (with severity prefixes) to \fI<filename>\fP in addition
+to stderr.
+.TP
+.BI -q|--quiet
+Suppress note-level and debug messages on stderr and in the log file.
 .TP
 .BI -t\ <filename>|--tools=<filename>
 Read Excellon tools from the file \fI<filename>\fP.
--- a/src/gerb_stats.c
+++ b/src/gerb_stats.c
@@ -318,6 +318,7 @@ gerbv_stats_add_error(gerbv_error_list_t *error_list_in,
             GERB_COMPILE_WARNING("%s",error_text);
             break;
         case GERBV_MESSAGE_NOTE:
+            GERB_NOTE("%s", error_text);
             break;
     }
 
--- a/src/gerbv.h
+++ b/src/gerbv.h
@@ -121,6 +121,7 @@ extern "C" {
 #define GERB_COMPILE_ERROR(...)  g_log(NULL, G_LOG_LEVEL_CRITICAL, __VA_ARGS__)
 #define GERB_COMPILE_WARNING(...)  g_log(NULL, G_LOG_LEVEL_WARNING, __VA_ARGS__)
 #define GERB_MESSAGE(...)  g_log(NULL, G_LOG_LEVEL_MESSAGE, __VA_ARGS__)
+#define GERB_NOTE(...)  g_log(NULL, G_LOG_LEVEL_INFO, __VA_ARGS__)
 
 /*! The aperture macro commands */  
 typedef enum {GERBV_OPCODE_NOP, /*!< no operation */
--- a/src/main.c
+++ b/src/main.c
@@ -137,6 +137,7 @@ const struct option longopts[] = {
     {"rotate",          required_argument,  NULL,    'r'},
     {"mirror",          required_argument,  NULL,    'm'},
     {"help",            no_argument,	    NULL,    'h'},
+    {"quiet",           no_argument,        NULL,    'q'},
     {"log",             required_argument,  NULL,    'l'},
     {"output",          required_argument,  NULL,    'o'},
     {"project",         required_argument,  NULL,    'p'},
@@ -162,7 +163,7 @@ const struct option longopts[] = {
     {0, 0, 0, 0},
 };
 #endif /* HAVE_GETOPT_LONG*/
-const char *opt_options = "VadhB:D:O:W:b:f:r:m:l:o:p:t:T:u:w:x:";
+const char *opt_options = "VadqhB:D:O:W:b:f:r:m:l:o:p:t:T:u:w:x:";
 
 /**Global state variable to keep track of what's happening on the screen.
    Declared extern in main.h
@@ -172,6 +173,8 @@ gerbv_screen_t screen;
 
 gboolean logToFileOption;
 gchar *logToFileFilename;
+static gboolean quietMode = FALSE;
+static FILE *logFile = NULL;
 
 /* Coords like "0x2" parsed by "%fx%f" will result in first number = 2 and
    second number 0. Replace 'x' in coordinate string with ';'*/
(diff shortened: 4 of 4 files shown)


spe-ciellt:
### Issues

**Fragile quiet-mode comparison.** GLib log levels are bitmasks, not an ordered scale — higher numeric values mean *less* severe:

```
G_LOG_LEVEL_ERROR    = 4
G_LOG_LEVEL_WARNING  = 16
G_LOG_LEVEL_INFO     = 64   ← what NOTE uses
G_LOG_LEVEL_DEBUG    = 128
```

The condition:
```c
if (quietMode && level >= G_LOG_LEVEL_INFO)
    goto handle_fatal;
```
happens to suppress INFO and DEBUG correctly by accident, but reads as if INFO is "high severity". It also silently suppresses `G_LOG_LEVEL_DEBUG` without any documentation. The correct idiom is an explicit bitmask:
```c
if (quietMode && (level & (G_LOG_LEVEL_INFO | G_LOG_LEVEL_DEBUG)))
    goto handle_fatal;
```

**`goto` violates codebase style.** `main.c` and the rest of the codebase don't use `goto`. The `handle_fatal:` label jump can be replaced with a plain `if`/`else` without any loss of clarity.

**One `exit(1)` path may not close the log file.** The `fclose(logFile)` calls are added before `exit(0)` and at GTK cleanup, but the `exit(1)` at "A valid file was not loaded" (after argument parsing, after the log file is already opened) appears to be missed. On most systems `exit()` flushes stdio anyway, but it is an inconsistency.

**"message" prefix may confuse users.** `GERB_MESSAGE()` uses `G_LOG_LEVEL_MESSAGE` which the PR labels `"message:"` on stderr. Most Unix CLI tools use `"info:"` for that severity. Minor naming issue but worth considering.

**`--quiet` suppresses notes during pass 1 of getopt only from pass 2 onward.** Any messages emitted during pass 1 of the two-pass `getopt` loop won't be suppressed because `quietMode` is set in pass 2. In practice almost no messages are emitted during option parsing, so this is unlikely to matter.

**Updating command line switches should update the man page as well.** I'm not sure if is up to date, but from now I think it is good idea to keep that updated as well.

rampageservices:
Thanks for the thorough review. All six points addressed:

1. **Fragile quiet-mode comparison** — Replaced `level >= G_LOG_LEVEL_INFO` with explicit bitmask `(level & (G_LOG_LEVEL_INFO | G_LOG_LEVEL_DEBUG))`.

2. **goto violates codebase style** — Replaced goto/label with an if/else block. The quiet-mode check now wraps the print logic in a negated condition, and the fatal handler runs unconditionally after.

3. **Missed exit(1) log file closure** — Added `if (logFile) fclose(logFile)` before both exit(1) paths at "A valid file was not loaded" (single-file export check and default export switch case).

4. **"message" prefix** — Changed `G_LOG_LEVEL_MESSAGE` prefix from `"message"` to `"info"`.

5. **--quiet only from pass 2** — Added `case 'q'` to the pass 1 getopt loop so `quietMode` is set before any messages can be emitted during pass 2 option parsing.

6. **Man page** — Added `-q|--quiet` entry and updated `-l|--log` description to reflect the new behavior (writes prefixed messages to file in addition to stderr).

Force-pushed with `--force-with-lease`.

rampageservices:
One concern: I'm uneasy about the naming where G_LOG_LEVEL_MESSAGE maps to "info" and G_LOG_LEVEL_INFO maps to "note". It feels inverted — "message" becoming the more informational label and "info" being demoted to "note" reads backwards. Open to suggestions on better naming here.

spe-ciellt:
I agree with you. You can swap them around if you want to. I don't know if that will cause any problems in other part of the code. If it is too many changes in other places of the code maybe we should leave it at the moment and drop an issue for future reference?

It would also be great if this could be documented somewhere. Currently I am a bit overwhelmed by all PRs coming in (I am positive, don't get me wrong). But keep this in mind for the future. ;)

rampageservices:
@spe-ciellt No urgency at all on any of our PRs — take your time. We really appreciate the decades of work you've put into gerbv, and the credit and attribution in the project. It's great to see a release happening after so long!

On the naming swap: I checked the scope — `GERB_MESSAGE` appears ~45 times across 10 files, `GERB_NOTE` is only in `gerb_stats.c`. So swapping them would be a mechanical rename but it touches a lot of call sites. I think the cleanest approach is to do the swap (rename `GERB_MESSAGE` → maps to `"note:"`, `GERB_NOTE` → maps to `"info:"`) in a separate follow-up PR so this one can land as-is. I'll drop an issue to track it and document the log level mapping there.

PR #293, opened 2026-02-28
fix: add bounds-check guards for file reading functions

rampageservices:
## Summary

`gerb_file_t` has two allocation paths in `gerb_fopen()`:

- **mmap path** (Linux/macOS): buffer is exactly `st_size` bytes with **no null terminator**
- **calloc path** (Windows): buffer has a null terminator at `data[datalen]`

`strtol()` and `strtod()` scan forward until they hit a non-matching character or `'\0'`. On the mmap path, if a file ends with digits and no trailing newline/command, these functions can read past the mapped region.

Additionally, `gerb_fgetdouble()` accesses `start[0]`/`start[1]` for the hex-float guard (added in #285) without checking that at least 2 bytes remain.

This PR:

- **Copies mmap'd data into a heap buffer with `'\0'` terminator** before `munmap`, making both code paths produce identical null-terminated buffers
- **Simplifies `gerb_fclose()`** — `fd->data` is now always heap-allocated, so the `#ifdef`/`munmap` branch is removed
- **Adds early-return guards** to `gerb_fgetint()` and `gerb_fgetdouble()` when the read pointer is at or past end-of-data
- **Guards the hex-float check** in `gerb_fgetdouble()` with `remaining >= 2` before accessing `start[1]`

## Affected functions

| Function | Fix |
|----------|-----|
| `gerb_fopen()` | Null-terminate mmap buffer via heap copy |
| `gerb_fclose()` | Remove `munmap` branch, use `g_free()` unconditionally |
| `gerb_fgetint()` | Early return at end-of-data |
| `gerb_fgetdouble()` | Early return at end-of-data + bounds-check hex guard |

Already safe (unchanged): `gerb_fgetc()` (checks `ptr >= datalen`), `gerb_fgetstring()` (uses `iend` bound).

## Testing

- Clean build, no new warnings
- 94/104 tests pass (same 10 pre-existing rendering mismatches, unrelated)
- Hex-expression test (`test-amacro-hex-expression.gbx`) renders correctly
- X2 attributes test (`test-gerber-x2-attributes.gbx`) renders correctly
- Truncated file ending in digits with no trailing newline — no crash

### DIFF
--- a/src/gerb_file.c
+++ b/src/gerb_file.c
@@ -102,12 +102,27 @@ gerb_fopen(char const * filename)
 
     dprintf("     Doing mmap\n");
     fd->datalen = (int)statinfo.st_size;
-    fd->data = (char *)mmap(0, statinfo.st_size, PROT_READ, MAP_PRIVATE, 
+    fd->data = (char *)mmap(0, statinfo.st_size, PROT_READ, MAP_PRIVATE,
 			    fd->fileno, 0);
     if(fd->data == MAP_FAILED) {
 	fclose(fd->fd);
 	g_free(fd);
 	fd = NULL;
+    } else {
+	/* Copy into a heap buffer with null terminator so strtol/strtod
+	 * have a safe stopping point — mmap does not guarantee '\0'
+	 * after the file content. */
+	char *buf = (char *)g_malloc(fd->datalen + 1);
+	if (buf == NULL) {
+	    munmap(fd->data, fd->datalen);
+	    fclose(fd->fd);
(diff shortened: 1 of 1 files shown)


spe-ciellt:
I merge this patch for now, looks simple and on point.

 This patch makes the Linux file operations and Windows file operation more similar, so they could probably be more "the same". The reason for selecting the mmap()/unmap() in the beginning was that it made parsing faster. So for Unix like system I would still prefer the mmap()/unmap().

PR #294, opened 2026-02-28
Fix all implicit case fallthrough (#287)

### LINKED ISSUE #287: Fix all case fallthrough
1. Annotate all cases which falls through with whatever GCC has decided is the latest for C. 
2. Enable the fall through checks in compilation.

rampageservices:
## Summary

- Enable `-Wimplicit-fallthrough` in `CMakeLists.txt` compiler flags
- Annotate all intentional case fallthroughs with C23 `[[fallthrough]];` attribute

## Changes

| File | Location | Fallthrough |
|------|----------|-------------|
| `src/amacro.c` | `case '0':` → `case '1':` | Comment is conditionally consumed, otherwise falls through to digit parsing |
| `src/callbacks.c` | `SAVE_FILE_RS274XM` → `SAVE_FILE_DRILL` | Merged visible layers export shares drill export path |
| `src/callbacks.c` | `LAYER_SELECTED` → `default` | Resolves selected index then falls to default toggle logic |
| `src/csv.c` (×2) | `ST_START` → `ST_COLLECT` | State-machine transition (char and wchar_t variants) |
| `src/drill.c` | `case 'R':` → `case 'S':` | Repeat-hole command falls through to spindle speed handler |
| `src/main.c` | `case 'w':` → `case 'W':` | Lowercase sets pixel flag then shares window-size parsing |

## Test plan

- [x] Clean build with zero `-Wimplicit-fallthrough` warnings
- [x] 94/104 regression tests pass (same 10 pre-existing rendering mismatches)
- [x] Functional test: `test-amacro-hex-expression.gbx` exports correctly
- [x] Functional test: `test-gerber-x2-attributes.gbx` exports correctly

Closes #287

### DIFF
--- a/src/drill.c
+++ b/src/drill.c
@@ -873,6 +873,7 @@ parse_drillfile(gerb_file_t *fd, gerbv_HID_Attribute *attr_list, int n_attr, int
 	      }
 	      
 	    }
+	    break;
 
 	case 'S':
 	    gerbv_stats_printf(stats->error_list, GERBV_MESSAGE_NOTE, -1,
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,7 +34,7 @@ set(SCMSUBDIR "scheme")
 set(BACKEND_DIR "${DATADIR}/${SCMSUBDIR}")
 
 add_library(compilation-flags INTERFACE)
-target_compile_options(compilation-flags INTERFACE -Wall -O3 -Wno-deprecated-declarations)
+target_compile_options(compilation-flags INTERFACE -Wall -Wimplicit-fallthrough -O3 -Wno-deprecated-declarations)
 target_compile_options(compilation-flags INTERFACE $<$<CONFIG:Debug>:-g3>)
 target_link_libraries(compilation-flags INTERFACE PkgConfig::GLIB PkgConfig::GTK m)
 if(DXF_FOUND)
--- a/src/amacro.c
+++ b/src/amacro.c
@@ -258,6 +258,7 @@ parse_aperture_macro(gerb_file_t *fd)
 		c = gerb_fgetc(fd); /* Read the '*' */
 		break;
 	    }
+	    [[fallthrough]];
 	case '1':
 	case '2':
 	case '3':
--- a/src/callbacks.c
+++ b/src/callbacks.c
@@ -697,6 +697,7 @@ callbacks_generic_save_activate (GtkMenuItem     *menuitem,
 			error_visible_layers = TRUE;
 			break;
 		}
+		break;
 	case CALLBACKS_SAVE_FILE_DRILL:
 		windowTitle = g_strdup_printf(
 			_("Export \"%s\" layer #%d to "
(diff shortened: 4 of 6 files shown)


src/drill.c
```diff
@@ -873,6 +873,7 @@ parse_drillfile(gerb_file_t *fd, gerbv_HID_Attribute *attr_list, int n_attr, int
 	      }
 	      
 	    }
+	    [[fallthrough]];
```
spe-ciellt: This is probably a bug and should be a `break`.  Remove the `[[fallthrough]]` line and replace the empty line with `break;`.

spe-ciellt (changes requested):
One old bug found, but was fixed in the wrong way.

spe-ciellt:
### Hard blocker — `drill.c` `case 'R'` → `case 'S'`

After processing a repeat-hole `R` command in the body section, the code falls through into `case 'S'` (spindle speed), which logs a spurious note and calls `eat_line(fd)` — consuming the **next line** from the drill file. That next line is typically a coordinate, tool change, or `M30` end-of-file command, which gets silently discarded.

```c
case 'R':
    if (state->curr_section == DRILL_HEADER) {
        // log error
    } else {
        // parse repeat count and step values
        // emit rcnt drill holes
    }
    // [[fallthrough]]; <-- PR annotates this as intentional
case 'S':
    gerbv_stats_printf(..., _("Ignoring setting spindle speed..."));
    eat_line(fd);
    break;
```

The existing golden-image test passes because it was generated against the same buggy code. The correct fix is a `break;` before `case 'S':`, not a `[[fallthrough]];`.

---

### Likely bug — `callbacks.c` RS274XM → DRILL

On the success path for `CALLBACKS_SAVE_FILE_RS274XM`, the code falls through into the `CALLBACKS_SAVE_FILE_DRILL` block which overwrites `windowTitle` (leaking the RS-274X string) with a drill-file dialog title and sets the file extension to `.drl`. An RS-274X merge export should not be presenting a drill-file dialog. This looks like a missing `break;` that has always been there, with the PR locking it in as intentional.

---

### Correct annotations (6 of 8)

| Location | Assessment |
|---|---|
| `amacro.c` — `case '0'` → `case '1'` | Correct — `'0'` mid-expression is a digit |
| `callbacks.c` — `LAYER_SELECTED` → `default` | Correct — replaces `/* No break */` which didn't match GCC's regex |
| `csv.c` (×2) — `ST_START` → `ST_COLLECT` | Correct — standard state machine optimization |
| `main.c` — `case 'w'` → `case 'W'` | Correct — intentional shared getopt handling |

The three existing `/* fall through */` comments in `scheme.c` are correctly left untouched — GCC level 3 already accepts them.

**When you fixed it**, amend it into current commit and push it `--force--with-lease`.

rampageservices:
Fixed both — `drill.c` case 'R' and `callbacks.c` RS274XM now have `break;` instead of `[[fallthrough]];`. Force-pushed the amended commit.

src/drill.c
rampageservices: This is now a proper break; (see below)

Issue #297, opened 2026-02-28
Excellon parser: missing G-code and M-code support


rampageservices:
## Summary

The Excellon drill parser (`src/drill.c`) defines many G-codes and M-codes in its enums (`drill_g_code_t`, `drill_m_code_t`) but only handles a subset. Commands that are parsed but not handled fall through to "unsupported" error messages, and commands not even parsed are silently ignored or reported as unknown.

This issue tracks the gap between what the parser defines and what it actually handles, prioritized by real-world impact.

### Handled
| Code | Name | Notes |
|------|------|-------|
| G00 | Rout mode / rapid positioning | Route mode state |
| G01 | Linear routing move | Route mode state (added in #283) |
| G02 | Circular CW arc routing | **Merged in #298** (I/J offsets), pending #345 (A radius) |
| G03 | Circular CCW arc routing | **Merged in #298** (I/J offsets), pending #345 (A radius) |
| G05 | Drill mode | Default mode |
| G32 | Routed circle CW canned cycle | Pending #305 |
| G33 | Routed circle CCW canned cycle | Pending #305 |
| G85 | Canned slot (drilled) | Parses end coords |
| G87 | Routed slot canned cycle | Pending #300 |
| G90 | Absolute mode | Coordinate mode |
| G91 | Incremental mode | Coordinate mode |
| G93 | Zero set | Origin offset |

### Handled (ignored — machine-only, no visual effect)
| Code | Name | Notes |
|------|------|-------|
| G07 | Override tool feed/speed | Pending #306 |
| G34-G39 | Vision system commands | Pending #306 |
| G40-G42 | Cutter compensation | Pending #306 |
| G45-G48 | Vision offset relative | Pending #306 |
| G81-G84 | Package canned cycles (DIP, etc.) | Pending #306 |

### Handled
| Code | Name | Notes |
|------|------|-------|
| M00 | End of program | |
| M01 | Pattern end | Pending #304 (currently stats-only) |
| M02 | Repeat pattern offset | Pending #304 |
| M14 | Z-axis route position (depth ctrl) | |
| M15 | Z-axis route position (tool down) | |
| M16 | Retract with clamping (tool up) | |
| M17 | Retract without clamping (tool up) | |
| M18 | Tool tip check | |
| M25 | Pattern start | Pending #304 (currently stats-only) |
| M30 | End + rewind | |
| M45 | Long message | |
| M47 | Message | |
| M48 | Header start | |
[...]

PR #298, opened 2026-02-28
fix: support Excellon G02/G03 arc routing with I/J center offsets

rampageservices:
## Summary

Add circular arc routing support to the Excellon drill parser:

- **G02** (clockwise) and **G03** (counter-clockwise) arc commands with **I/J center offsets** are now parsed and rendered
- Reuses the existing multi-quadrant arc algorithm (`calc_cirseg_mq` from `gerber.c`) for consistent arc geometry
- Adds I/J offset parsing to the Excellon coordinate handler

## Why

Modern CAD tools (KiCad, Altium, etc.) emit G02/G03 arc routing commands for arc-routed slots, rounded board edges, and curved milling paths. Previously these commands were silently ignored, causing incomplete rendering of routed features.

## How

- New `drill_add_arc_segment()` function in `src/drill.c` that constructs a `gerbv_cirseg_t` from the start point, end point, and I/J center offsets, computing arc angles via the same quadrant logic as Gerber circular interpolation
- Extended the Excellon coordinate parser to recognize `I` and `J` offset tokens and store them in `drill_state_t`
- G02/G03 commands (with tool down in route mode) now dispatch to `drill_add_arc_segment()` instead of being ignored

## Test

New test file `test/inputs/test-drill-arc-routing.exc` covering:
- CW 90° arc (G02)
- CCW 180° arc (G03)
- Full 360° circle (G02)

Golden reference image at `test/golden/test-drill-arc-routing.png`. Test registered in `test/tests.list`. All pre-existing tests continue to pass (95/105 pass; 10 pre-existing failures unrelated to this change).


rampageservices:
Addresses the highest-priority item in #297 (G02/G03 arc routing).

rampageservices:
**Upstream dependencies:**

- This PR depends on #283 (Excellon route milling), which is already merged — it provides `drill_add_route_segment()`, the `route_mode`/`tool_down` state tracking, and the M15/M16/M17 tool up/down handling that this PR builds on for arc routing.
- This PR may have minor merge conflicts with #300 (G87 routed slot) in the G-code dispatch switch in `drill.c`, since both add cases in the same region. Whichever merges second will need a trivial rebase.

spe-ciellt:
Yes, I will probably need information in what order you have thought all your PRs should be merged? I will check all your PRs in the coming days and drop comments.

spe-ciellt:
Comments for future updates

### 1. Code duplication — `calc_cirseg_mq` and `calc_cirseg_bbox` re-implemented verbatim

Both are `static` in `gerber.c` so they can't be called from `drill.c`. The PR duplicates:
- Full body of `calc_cirseg_mq()` (~30 lines of math + comments)
- Most of `calc_cirseg_bbox()` (~25 lines) reimplemented inline in the bbox block

If a bug is ever found in `calc_cirseg_mq`, it needs fixing in two places. The correct structural fix is to move both functions to a shared location (e.g., `gerb_image.c`) and export them. That's a larger refactor — acceptable to defer, but worth a TODO comment or a follow-up issue.

### 2. Bounding box uses `floor()` — differs from `calc_cirseg_bbox` but is actually more correct

The PR's middle-point loop:
```c
for (step_pi_2 = (floor(ang1 / M_PI_2) + 1) * M_PI_2; ...)
```
`calc_cirseg_bbox` in `gerber.c` uses:
```c
for (step_pi_2 = (ang1/M_PI_2 + 1)*M_PI_2; ...)
```
For non-axis-aligned `ang1` (e.g. 45°), `gerber.c` gives start = `1.5×(π/2)` = 135° — which **skips the 90° axis crossing**. The PR's `floor(0.5)=0` gives start = `1×(π/2)` = 90° — which is **correct** (first axis crossing after 45°). The PR accidentally fixes a latent imprecision in the original bbox algorithm. No action needed here, but it's worth documenting.

### 3. `delta_cp_x/y` not reset on T-code tool change (`drill.c:887`)

```c
case 'T':
    drill_parse_T_code(fd, state, image, file_line);
    state->route_mode = DRILL_G_DRILL;
    state->tool_down = FALSE;
    /* delta_cp_x/y NOT reset here */
    break;
```
In practice this is harmless — both fields are cleared after every arc segment and on tool-up, so by the time a T-code is seen they're already 0. But for defensive consistency with the `route_mode`/`tool_down` resets, adding `state->delta_cp_x = state->delta_cp_y = 0;` here would be better hygiene.

Great patch, will merge it and drop an issue for future reference.

PR #300, opened 2026-02-28
feat: implement Excellon G87 routed slot canned cycle

rampageservices:
## Summary

- Implement parsing and rendering of the Excellon G87 routed slot canned cycle
- G87 works like G85 (slot from current XY to end XY) but creates a **routed** line segment via `drill_add_route_segment()` instead of modifying an existing drill flash
- Some CAD tools (e.g. certain Altium and KiCad exporters) emit G87 instead of G85 for routed slots; without this patch they fall through to "Unsupported G-code"

## Changes

- **`src/drill.c`** — Parse `case 87:` in `drill_parse_G_code()` stats; add `case DRILL_G_ROUTSLOT:` handler that reads end-point coordinates and calls `drill_add_route_segment()`
- **`src/gerbv.h`** — Add `int G87` counter to `gerbv_drill_stats_t`
- **`src/drill_stats.c`** — Accumulate G87 stats across layers
- **`src/callbacks.c`** — Display G87 row in drill analysis stats UI
- **`test/inputs/test-drill-routed-slot.exc`** — Test file with horizontal and vertical G87 routed slots using two tool sizes
- **`test/golden/test-drill-routed-slot.png`** — Golden reference image

## Test plan

- [ ] `cmake --build build` compiles cleanly with no warnings
- [ ] `cd build && ctest` — new `test-drill-routed-slot` passes, no existing test regressions
- [ ] Visual check: `./build/src/gerbv test/inputs/test-drill-routed-slot.exc` renders two routed slots (one horizontal, one vertical)


rampageservices:
**Upstream dependencies:**

- This PR depends on #283 (Excellon route milling), which is already merged — it provides `drill_add_route_segment()` used by the G87 handler.
- This PR may have minor merge conflicts with #298 (G02/G03 arc routing) in the G-code dispatch switch in `drill.c`, since both add cases in the same region. Whichever merges second will need a trivial rebase.

spe-ciellt:
### Test file format inconsistency — first coordinate pair is silently discarded

The G85 (drilled slot) test file uses the pattern:
```
X2.0872Y0.6494G85X2.1463Y0.6494   ← pre-position, then G85 end only
```
The PR description states G87 "works like G85 (slot from current XY to end XY)". Consistent with that, the correct G87 test file format would be:
```
X10000Y10000          ← rapid move to start
G87X20000Y10000       ← route to end
```

But the test file uses:
```
G87X10000Y10000X20000Y10000
```

Inside `drill_parse_coordinate` the loop reads all X/Y tokens with last-write-wins. It processes `X10000`, `Y10000`, then `X20000` (overwrites), `Y10000` — ending with `curr_x = 20000-format`, `curr_y = 10000-format`. The first pair (`X10000Y10000`) is silently discarded. Meanwhile `prev_x/prev_y = state->curr_x = 0` (tool was at origin, never positioned).

Result: the slot runs from `(0, 0)` to `(20mm, 10mm)` — a **diagonal**, not the "horizontal slot" the PR description claims. The golden image was generated from this behavior, so the test passes, but it is testing the wrong geometry.

For a horizontal slot from (10mm, 10mm) to (20mm, 10mm), the test file should either:
- Use `X10000Y10000` on its own line to pre-position, then `G87X20000Y10000`
- OR the implementation needs to split the two pairs explicitly (if G87's format genuinely embeds both start and end inline)

rampageservices:
Good catch @spe-ciellt — you're right, the test file had both coordinate pairs inline on the G87 line, so `drill_parse_coordinate` last-write-wins discarded the first pair. The slot was running from `(0,0)` to `(20mm,10mm)` instead of `(10mm,10mm)` to `(20mm,10mm)`.

Fixed in 755fab7 — split into separate pre-position + G87 lines, matching the G85 drilled slot test pattern:

```
T01
X10000Y10000
G87X20000Y10000
T02
X15000Y5000
G87X15000Y15000
```

The G87 implementation itself is correct — it routes from current position to the specified end coordinate. The test file was wrong, not the code. Golden image regenerated; now shows the intended horizontal + vertical cross.

This PR also needs a rebase onto current `develop` — I'll do that next.

PR #303, opened 2026-03-01
cmake: fix tarball builds and generated path handling

chenrui333:
Fix CMake tarball builds used by downstream packaging.

- fall back to `PROJECT_VERSION` when Git metadata is unavailable
- fix generated-source path handling for gettext inputs
- avoid configure-time manpage compression failure by installing the configured manpage directly
- keep Homebrew-specific compile/rpath adjustments in-tree for macOS/Linux

Ref: https://github.com/Homebrew/homebrew-core/pull/269931

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,6 +43,12 @@ endif()
 if(Intl_FOUND)
     target_link_libraries(compilation-flags INTERFACE Intl::Intl)
 endif()
+if(APPLE)
+    target_compile_definitions(compilation-flags INTERFACE QUARTZ)
+endif()
+if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
+    target_compile_definitions(compilation-flags INTERFACE _GNU_SOURCE)
+endif()
 
 ## This is installation directory settings. These must be solved later by install target
 target_compile_definitions(compilation-flags INTERFACE "GERBV_DATADIR=\"${DATADIR}\"")
--- a/cmake/GetGitVersion.cmake
+++ b/cmake/GetGitVersion.cmake
@@ -18,6 +18,11 @@ if(__get_git_version)
 endif()
 set(__get_git_version INCLUDED)
 function(get_git_version var)
+  set(FALLBACK_GIT_VERSION "v0.0.0")
+  if(PROJECT_VERSION)
+      set(FALLBACK_GIT_VERSION "v${PROJECT_VERSION}")
+  endif()
+
   if(GIT_EXECUTABLE)
       execute_process(COMMAND ${GIT_EXECUTABLE} describe --tag --match "v[0-9]*.[0-9]*.[0-9]*" --abbrev=8
           WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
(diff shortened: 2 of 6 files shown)


spe-ciellt:
Thank you for this pull request, looks relevant. I will double check this and merge if OK in a couple of days.

rampageservices:
Reviewed for security and consistency. Looks good overall — no security concerns, sensible packaging fixes. A few notes:

**GetGitVersion.cmake** — The `PROJECT_VERSION` fallback and moving the dirty check inside the `else` (successful git describe) branch are both correct. Previously, a failed `git describe` would still run the dirty check against a potentially absent HEAD, which could produce confusing errors in tarball builds.

**config/CMakeLists.txt** — Changing `generated` from `add_library(INTERFACE)` to `add_custom_target(DEPENDS ...)` is the right call. The old `target_sources(generated PRIVATE bugs.c authors.c)` on an INTERFACE library is questionable CMake — INTERFACE libraries shouldn't own source files. A custom target purely expressing the dependency is cleaner.

**locale/CMakeLists.txt** — Using `get_filename_component()` to resolve absolute path and adding `add_dependencies(${GETTEXT_TARGET} generated)` fixes the build order issue where gettext might try to scan `authors.c`/`bugs.c` before they're generated. Correct.

**man/CMakeLists.txt** — Removing the `file(ARCHIVE_CREATE)` gzip step and installing the uncompressed `.1` file directly. This delegates compression to the packaging system (dpkg, rpm, etc.) which is the standard approach. Clean.

**src/CMakeLists.txt** — Replacing `target_link_libraries(... generated)` with `add_dependencies(gerbv-app generated)` is consistent with the config/ change. The macOS INSTALL_RPATH addition is correct for Homebrew's lib layout.

**CMakeLists.txt** — `QUARTZ` define on APPLE is needed — `callbacks.c` and `render.c` use `#if defined(QUARTZ)` for macOS-specific paths, and the CMake build was not defining it (commit ae525ba changed the checks from `#if QUARTZ` to `#if defined(QUARTZ)` in anticipation of this). `_GNU_SOURCE` on Linux is used by `lrealpath.c` for `canonicalize_file_name`.

No issues found. LGTM.

rampageservices:
**Additional note:** the `_GNU_SOURCE` define on Linux also fixes pre-existing undeclared-identifier errors for `M_PI_2` and `M_1_PI`.

These are POSIX/XSI math constants (not C standard), and the project sets `CMAKE_C_EXTENSIONS OFF` which produces `-std=c23` instead of `-std=gnu23`. Without a feature-test macro, `<math.h>` on glibc won't expose them.

| Constant | Where used |
|----------|------------|
| `M_PI_2` | `draw-gdk.c`, `draw.c`, `gerb_image.c`, `interface.c`, `gerber.c` (15 uses total) |
| `M_1_PI` | `gerbv.h:118` via `RAD2DEG()` — propagates to every file including the public header |
| `M_PI`   | `gerbv.h:117` via `DEG2RAD()`, `draw-gdk.c` |

`_GNU_SOURCE` implies `_XOPEN_SOURCE` on glibc, which gates all three constants in `<math.h>`. On macOS they're available by default, so the Linux-only scope is correct. This PR effectively fixes the build on strict-C toolchains.

chenrui333:
Thanks for the merge, also log a pr to add macos CI, https://github.com/gerbv/gerbv/pull/327.

PR #307, opened 2026-03-01
feat: warn when loaded file contains no geometry

### LINKED ISSUE #259: feature request: report / indicate empty gerber files
It would be useful if gerbv had the ability to report / indicate that some of the gerber files loaded are empty. "Empty" in this context means that the file has correct syntax, but defines no actual features. Example: a drill file that has no drill holes included.

Background: some tools generate empty gerber files (yes, it is a bug). Often gerbv is used to check the generated output, so getting a "heads up" about empty files would be nice.

rampageservices:
## Summary

Closes #259.

- Adds a post-parse emptiness check in `gerbv_open_image()` that emits a `GERBV_MESSAGE_WARNING` when a file parses successfully but contains no actual geometry (the net list has only the sentinel node, no real entries)
- Drill files get: `File "..." contains no drill hits or routes` (via `drill_stats->error_list`)
- Gerber files get: `File "..." contains no drawing elements` (via `gerbv_stats->error_list`)
- Single insertion point in `src/gerbv.c`, no changes to parsers

## Test plan

- [x] `cmake --build build` — clean build, zero new warnings
- [x] `ctest` — 94/104 pass (10 pre-existing image mismatches, unchanged)
- [x] `test/inputs/empty.xnc` (existing empty drill file) produces the drill warning
- [x] Minimal empty Gerber file (headers + M02, no draws) produces the Gerber warning
- [x] Non-empty files produce no false positive warnings

### DIFF
--- a/src/gerbv.c
+++ b/src/gerbv.c
@@ -602,10 +602,10 @@ gerbv_open_image(gerbv_project_t *gerbvProject, gchar const* filename, int idx, 
     
     if (parsed_image) {
 	/* Warn if the file parsed successfully but contains no geometry */
-	if (parsed_image->netlist != NULL
-	&&  parsed_image->netlist->next == NULL) {
-	    if (parsed_image->layertype == GERBV_LAYERTYPE_DRILL
-	    &&  parsed_image->drill_stats != NULL) {
+	if ((parsed_image->netlist != NULL)
+	&&  (parsed_image->netlist->next == NULL)) {
+	    if ((parsed_image->layertype == GERBV_LAYERTYPE_DRILL)
+	    &&  (parsed_image->drill_stats != NULL)) {
 		gerbv_stats_printf(
 		    parsed_image->drill_stats->error_list,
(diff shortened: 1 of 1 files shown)


src/gerbv.c
```diff
@@ -600,6 +601,25 @@ gerbv_open_image(gerbv_project_t *gerbvProject, gchar const* filename, int idx,
     }
     
     if (parsed_image) {
+	/* Warn if the file parsed successfully but contains no geometry */
+	if (parsed_image->netlist != NULL
```
spe-ciellt: Could you change this to surround every comparison with parenthesis?
Like `((a == b) && (c == d))` while there are boolean arithmetic with `&&` and `||` . There are several places in this patch.
I know I am bit inconsequent, but that is how I like it to be. I really don't trust precedence.

spe-ciellt (commented):
Only one thing but in several places.

rampageservices:
@spe-ciellt Added parentheses around comparisons in boolean expressions as requested — see latest commit 391db5a.

PR #312, opened 2026-03-01
docs: audit and update man page for current codebase

rampageservices:
## Summary

The man page (`gerbv.1.in`) was last updated July 2013 and had several discrepancies with the current codebase. This PR brings it up to date.

### Changes

- **Fix header date typo**: "Jule 13, 2013" (typo + stale)
- **Remove dead link**: `http://pcb.geda-project.org/` is no longer active
- **Add missing CLI options**: `-r, --rotate` and `-m, --mirror` are implemented in main.c but were never documented in the man page
- **Update export format list**: the man page listed `png/pdf/ps/svg/rs274x/drill` but `idrill` (ISEL NCP drill) is always available and `dxf` is conditionally available when compiled with libdxflib
- **Remove outdated scroll wheel instructions**: "if you enabled that in your X server and mapped it to button 4 and 5" hasn't been relevant in 15+ years
- **Update SUPERIMPOSING section**: described only legacy GDK compositing modes (GDK_COPY, GDK_XOR, etc.) without mentioning Cairo, which has been the default renderer since ~2.7. Updated to describe Cairo's alpha transparency behavior as the primary mode, with GDK modes noted as legacy

### Not changed (questions for @spe-ciellt)

- **Copyright years** (2001-2008) and **AUTHOR section**: Left as-is. The existing `CONTRIBUTORS` file tracks contributors — should the man page copyright years be extended, or is the current range intentional (covering the original authorship period)?
- **`-d, --dump` option**: Intentionally undocumented (debug flag, not in `--help` output either)

## Test plan

- [x] `cmake --build build` — clean build
- [x] `man -l build/man/gerbv.1` — renders correctly, no formatting issues

### DIFF
--- a/man/gerbv.1.in
+++ b/man/gerbv.1.in
@@ -1,4 +1,4 @@
-.TH gerbv 1 "Jule 13, 2013" Version @VERSION@
+.TH gerbv 1 "March 2026" Version @VERSION@
 .SH NAME
 gerbv - Gerber Viewer
 .SH SYNOPSIS
@@ -9,7 +9,7 @@ gerbv - Gerber Viewer
 RS274-X files are generated from different PCB CAD programs and are
 used in the printed circuit board manufacturing process.
 \fIgerbv\fP also supports Excellon/NC drill files as well as XY (centroid)
-files produced by the program PCB (http://pcb.geda-project.org/).
+pick-and-place files.
 
 .SH OPTIONS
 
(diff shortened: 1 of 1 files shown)


spe-ciellt:
Great question @rampageservices .

Never thought of who the AUTHORS and COPYRIGHT was in those man pages. And that it was written like that.

For copyright you can probably set me 2001-2026 if you think it sounds OK. For the Authors we should maybe create a THANKS file as `curl` have?

But I will merge this now so it is more correct regarding the other stuff.

PR #319, opened 2026-03-01
Fix errno clobbered by cleanup in gerb_fopen()

rampageservices:
## Summary

- Preserve `errno` across cleanup calls (`fclose`, `g_free`, `munmap`) in all error paths of `gerb_fopen()`
- Fix a latent NULL pointer dereference on the `mmap` failure path

Found while reviewing #302 — the `strerror(errno)` pattern used by all three callers (`gerbv_open_image`, `gerbv_create_rs274x_image_from_filename`, `gerbv_create_excellon_image_from_filename`) relies on `gerb_fopen()` preserving `errno`, but the cleanup calls before `return NULL` could clobber it.

## Details

`gerb_fopen()` has six error return paths after the initial allocation. In each case, cleanup functions (`fclose()`, `g_free()`, `munmap()`) run between the failing call and `return NULL`. These cleanup functions are allowed to modify `errno`, so callers that use `strerror(errno)` may get an unrelated error message.

The fix uses the standard POSIX `saved_errno` idiom — save `errno` immediately after the failing call, perform cleanup, then restore it before returning. This is the established pattern in GLib/POSIX C code (used throughout glibc internals, libgit2, curl, and GLib itself) for any wrapper function that performs cleanup between a failing syscall and its return.

Additionally, the `mmap` failure path previously set `fd = NULL` and fell through to `fd->filename = g_strdup(...)`, which would dereference NULL. This is now a proper `return NULL`.

## Affected error paths

| Failing call | Cleanup before return | errno preserved |
|---|---|---|
| `g_fopen()` | `g_free()` | now yes |
| `fstat()` | `fclose()`, `g_free()` | now yes |
| `mmap()` | `fclose()`, `g_free()` | now yes (+ now returns NULL) |
| `g_malloc()` | `munmap()`, `fclose()`, `g_free()` | now yes |
| `calloc()` | `fclose()`, `g_free()` | now yes |
| `fread()` | `fclose()`, `g_free()` x2 | now yes |

## Test plan

- [x] Clean build, zero warnings
- [x] 94/104 regression tests pass (10 pre-existing image comparison mismatches)

Ref: #302

### DIFF
--- a/src/gerb_file.c
+++ b/src/gerb_file.c
@@ -69,7 +69,9 @@ gerb_fopen(char const * filename)
     /* fopen() can't open files with non ASCII filenames on windows */
     fd->fd = g_fopen(filename, "rb");
     if (fd->fd == NULL) {
+	int saved_errno = errno;
 	g_free(fd);
+	errno = saved_errno;
 	return NULL;
     }
 
@@ -77,8 +79,10 @@ gerb_fopen(char const * filename)
     fd->ptr = 0;
     fd->fileno = fileno(fd->fd);
     if (fstat(fd->fileno, &statinfo) < 0) {
+	int saved_errno = errno;
 	fclose(fd->fd);
 	g_free(fd);
+	errno = saved_errno;
 	return NULL;
     }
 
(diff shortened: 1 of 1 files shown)


spe-ciellt:
`g_malloc()` NULL path is dead code — `saved_errno` there is never reached

GLib's `g_malloc()` is documented to never return NULL: on allocation failure it calls `g_error()` (which terminates the process). The `if (buf == NULL)` block at line 116 is therefore dead code, and the `saved_errno` additions there:

```c
char *buf = (char *)g_malloc(fd->datalen + 1);
if (buf == NULL) {          /* ← can never be true */
    int saved_errno = errno;
    munmap(fd->data, fd->datalen);
    fclose(fd->fd);
    g_free(fd);
    errno = saved_errno;
    return NULL;
}
```

are unreachable. The PR did not introduce the dead `if (buf == NULL)` guard (it's pre-existing), and applying the idiom uniformly is defensively correct, so this is not a blocker. However, it's worth a comment that `g_malloc` is abort-on-OOM in GLib.

Similarly, `g_new()` at the top of the function (`fd = g_new(gerb_file_t, 1)`) never returns NULL, making the `if (fd == NULL) return NULL` check at line 64 also unreachable. Same pre-existing issue, not introduced here.

**Considered minor and save for future updates.**

PR #323, opened 2026-03-01
fix: unwind file pointer on mismatch in drill_parse_header_is_metric()

rampageservices:
## Summary

- Unrolls compound `gerb_fgetc()` calls in `drill_parse_header_is_metric()` into separate checks with proper `gerb_ungetc()` on every failure path
- Previously, when multi-byte sequences failed to match partway through (e.g. "METRIC" vs "M48"), consumed bytes were not pushed back, causing the next parser to read from the wrong position
- Also improves FMT_000_000 vs FMT_000_00 final digit handling with block-scoped variable and EOF check

## Test plan

- [x] `make` builds cleanly with no warnings
- [x] `make check` — 93 passed, 11 failed (matches baseline, all failures are pre-existing)

Port of Henry Gabryjelski's commit 71604efc, adapted for current develop.

### DIFF
--- a/src/drill.c
+++ b/src/drill.c
@@ -1848,16 +1848,24 @@ drill_parse_header_is_metric(gerb_file_t *fd, drill_state_t *state,
 		break;
 	    }
 
-	    if ('0' == gerb_fgetc(fd)
-	    &&  state->autod) {
-		state->number_format = FMT_000_000;
-		state->decimals = 3;
-	    } else {
-		gerb_ungetc(fd);
-
-		if (state->autod) {
-		    state->number_format = FMT_000_00;
-		    state->decimals = 2;
+	    /* Either FMT_000_000 or FMT_000_00.
+	     * No longer rewind to beginning, as either option
+	     * results in an accepted value. */
+	    {
+		int last_char = gerb_fgetc(fd);
+		if (last_char == '0') {
+		    if (state->autod) {
+			state->number_format = FMT_000_000;
+			state->decimals = 3;
+		    }
+		} else {
+		    if (last_char != EOF) {
+			gerb_ungetc(fd);
+		    }
+		    if (state->autod) {
+			state->number_format = FMT_000_00;
+			state->decimals = 2;
+		    }
 		}
 	    }
 


spe-ciellt:
@rampageservices Generally a  good idea to merge this if it is still relevant, but it currently have merge conflicts that must be resolved. I don't know if these conflicts are caused by an already similar fix or if something has happened.
Thank you!

spe-ciellt:
## Issues Found

### 1. `state->autod` guards are already in current develop — conflict source

The diff shows these as new additions in both `drill_parse_header_is_metric()`
and `drill_parse_header_is_inch()`:

```diff
-    state->unit = GERBV_UNIT_MM;
+    if (state->autod) {
+        state->unit = GERBV_UNIT_MM;
+    }
```

The current develop tree already has both guards applied:

```c
/* current develop — drill.c */
if (state->autod) {
    state->unit = GERBV_UNIT_MM;
}
return 1;
} /* drill_parse_header_is_metric() */
```

and identically for `GERBV_UNIT_INCH`. These landed independently of this
PR and are the primary source of the dirty/conflicting status. A rebase must
drop these hunks since they are already present.

### 2. `goto header_junk` conflicts with `found_junk` refactoring in develop

The PR introduces per-path `gerb_ungetc()` sequences followed by
`goto header_junk`. The current develop version of the same function has
been refactored to use a `found_junk` boolean with a centralized handler:

```c
/* current develop */
gboolean found_junk = FALSE;
...
found_junk = TRUE;
break;
...
if (found_junk) {
    gerb_ungetc(fd);
    eat_line(fd);
    gerbv_stats_printf(...warning...);
}
```

There is no `goto header_junk` label in the current function. The PR's
`goto` hunks will conflict directly with this structure.

Importantly, the `found_junk` + single `gerb_ungetc()` + `eat_line()`
approach is functionally correct for error recovery: `eat_line()` skips all
remaining characters to the end of the current line regardless of how many
bytes were consumed before the error. The final file position — "start of
next line" — is the same whether one or N ungets are performed before
`eat_line()`. The per-path unget counting in the PR provides no additional
correctness benefit over the current pattern.

The rebase should keep `found_junk` and extract only the EOF guard fix (see
"What's Correct") as a targeted change.

### 3. Compound `||` check in `FMT_0000_00` path left unsplit — ungetc count is wrong

The `FMT_0000_00` two-trailing-zeros check is kept compound in the PR:

```c
if ('0' != gerb_fgetc(fd)
||  '0' != gerb_fgetc(fd)) {
    /* read four extra characters that didn't match */
    gerb_ungetc(fd);
    gerb_ungetc(fd);
    gerb_ungetc(fd);
    gerb_ungetc(fd);
    goto header_junk;
}
```

Due to C short-circuit evaluation:
- If the **first** call returns non-`'0'`: second is not called → 1 byte
  consumed here (plus 5 before = 6 total). 4 pre-goto ungets + 1 at
  `header_junk:` = 5 total, leaving the pointer 1 byte short.
- If the **first** returns `'0'` but **second** does not: 2 bytes consumed
  here (7 total). 5 total ungets, still 2 bytes short.

In both cases the rewind lands within the already-scanned portion of the
current line, so `eat_line()` correctly finishes at the line boundary.
The behaviour is harmless but the comment "read **four** extra characters"
is inaccurate in both sub-cases, and this check was not unrolled consistently
with the other checks in the same case block.

### 4. No regression test for the EOF double-read bug

The EOF-before-`gerb_ungetc` bug can only be triggered by a drill file whose
`METRIC,000.00` field is terminated by EOF rather than a newline — unusual
but possible. Adding a minimal test fixture (`test-drill-metric-eof.drl`)
that ends the `METRIC,000.00` format field with EOF and a corresponding entry
in `test/tests.list` would lock in the fix.

---

## Overall Verdict

The PR carries one genuine bug fix that is not in develop: the EOF guard
before `gerb_ungetc()` in the FMT_000_000 vs FMT_000_00 disambiguation,
which prevents a silent pointer-undercount when a drill file ends mid-format.
The `state->autod` guards are already in develop and are the primary conflict
source. The `goto header_junk` unwind approach conflicts with the `found_junk`
refactoring that replaced it, and provides no correctness benefit over the
existing `eat_line()` pattern. The right rebase strategy is to drop the
already-landed autod changes, keep `found_junk` control flow, and extract
only the `if (last_char != EOF)` guard as a two-to-four-line targeted fix.

| Severity     | Issue |
|--------------|-------|
| **Blocking** | `state->autod` guards already in develop — rebase conflict |
| **Blocking** | `goto header_junk` conflicts with `found_junk` refactoring in develop |
| Maintenance  | `FMT_0000_00` compound `||` check unsplit; ungetc count inaccurate (harmless) |
| Maintenance  | No test for EOF-terminated format field |

Needs rebase before merge. The extractable fix — `if (last_char != EOF)` guard
before `gerb_ungetc()` — is genuinely needed and should be preserved.

rampageservices:
@spe-ciellt Thanks for the thorough review. Rebased in f06f48c — the resolution addresses all your points:

1. **autod guards** — dropped; already in develop via #313.
2. **`goto header_junk`** — replaced with the existing `found_junk` pattern. No goto label, no per-path ungetc sequences.
3. **EOF guard** — preserved the `if (last_char != EOF) gerb_ungetc(fd)` check in the FMT_000_000/FMT_000_00 disambiguation, which is the one genuine fix not in develop.
4. **Test for EOF-terminated format field** — fair point, I'll add one.

The net diff against develop is now just the restructured FMT_000_000/FMT_000_00 block with the EOF guard.

PR #327, opened 2026-03-01
ci: add macos test

chenrui333:
Add a native macOS CI smoke build on `macos-15` for CMake/Homebrew validation.

Follow-up to #303.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -14,7 +14,7 @@ on:
 
 
 jobs:
-  ci:
+  ci-ubuntu-22.04:
     runs-on: ubuntu-22.04
     strategy:
       matrix:
@@ -22,7 +22,6 @@ jobs:
         include:
           - coverage: ""
             code_coverage: "" # Always empty, so matrix.code_coverage is always false
-    name: ci
 
     steps:
     - uses: actions/checkout@v6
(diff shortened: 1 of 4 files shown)


spe-ciellt:
Thank you for the MacOS support för building.

I would ask if you possible could create a preset and a toolchain file for MacOS in `cmake/preset` and `cmake/toolchain`. And then use it in the script. If not here, we could create an issue about it and fix it later.

Also, you dont have to refer to previous commit in code.

chenrui333:
thanks for the feedback, I am updating the pr now.

chenrui333:
Followed up on this in `b27005a`.

Changes made:
- Added a dedicated macOS preset: `cmake/preset/macos-clang.json`
- Added macOS toolchain file: `cmake/toolchains/macos-clang.cmake`
- Included the new preset in `CMakePresets.json`
- Updated macOS CI steps to use presets directly:
  - `cmake --preset macos-clang`
  - `cmake --build --preset macos-clang --parallel ...`
  - `ctest --preset macos-clang -V`

I also set `CMAKE_COMPILE_WARNING_AS_ERROR=OFF` in the macOS preset so the smoke job does not fail on existing Clang-only warnings unrelated to this CI wiring.

rampageservices (commented):
A few observations from reviewing:

**`_GNU_SOURCE` in `cmake/toolchains/macos-clang.cmake`** — consistent with the Linux toolchain which also defines it, so this makes sense for parity.

**`continue-on-error: true` on the test step** — is this intentional as a temporary measure while macOS tests stabilize, or should it stay permanently? If temporary, it might be worth adding a comment noting why, so it doesn't get forgotten. If permanent, it means macOS test failures will never block CI, which reduces the value of running them.

**`CMAKE_COMPILE_WARNING_AS_ERROR: OFF`** — the Linux preset doesn't set this at all (inherits whatever the project default is). Is macOS hitting warnings that Linux doesn't? If so, it might be worth documenting which warnings are macOS-specific, since silencing them project-wide could mask real issues on that platform.

**bzip2 pkg-config file generation** — this is the standard approach for Homebrew keg-only packages that don't ship .pc files. Looks correct.

**Job rename from `ci` to `ci-ubuntu-22.04`** — worth double-checking that no branch protection rules reference the old `ci` status check name, since GitHub treats these as distinct checks.

**`actions/checkout@v6`** — consistent with the existing Ubuntu job. Neither is SHA-pinned, which is fine for a project of this size, but noting it for completeness.

Overall looks solid as a smoke-test addition. The main question is whether `continue-on-error` should remain long-term.

spe-ciellt:
Thank you for the `cmake/preset` and `cmake/toolchains` files. Now we can signal that we support MacOS as well.

I just checked the CI build and it reported the following error https://github.com/gerbv/gerbv/actions/runs/22560526941/workflow

Seems there can not be number in ci work names. And also the comnment above by @rampageservices on at least commenting if it is temporary or permanently off.

It is great to test it on Clang as well and (later) making sure it  is warning free there as well. But it is a journey...

chenrui333:
yeah, that should easy fix (I ran into that issue the other day, that the job id does not support dot in `22.04`

chenrui333:
@spe-ciellt should be good now

spe-ciellt:
Thanks for the updates. 

OK, I see the warnings from Clang now.  Interesting, Clang and GCC seems to disagree how much `[[fallthrough]]` to add, This is something to analyze.

I will merge this now.

chenrui333:
Thanks @spe-ciellt.

PR #331, opened 2026-03-02
Parse Allegro comment-style tool definitions in Excellon drill files

### LINKED ISSUE #96: Incorrect rendering for Allegro Excellon files with "comment-style" tooldefs
gerbv does not parse the type of tool definition Allegro exports, and will render files using these tool definitions using some (incorrect) built-in defaults. Example file: [drill.zip](https://github.com/gerbv/gerbv/files/8009950/drill.zip)

Partial log:
```
** (process:857403): CRITICAL **: 11:35:53.848: Tool 01 used without being defined at line 22 in file "/home/jaseg/proj/gerbolyze/gerbonara/gerbonara/tests/resources/allegro-2/MinnowMax_RevA1_DRILL/MinnowMax_RevA1_NCDRILL.drl"
** (process:857403): WARNING **: 11:35:53.850: Setting a default size of 0.024"
** (process:857403): CRITICAL **: 11:35:53.851: Tool 02 used without being defined at line 1853 in file "/home/jaseg/proj/gerbolyze/gerbonara/gerbonara/tests/resources/allegro-2/MinnowMax_RevA1_DRILL/MinnowMax_RevA1_NCDRILL.drl"
** (process:857403): WARNING **: 11:35:53.851: Setting a default size of 0.032"
** (process:857403): CRITICAL **: 11:35:53.851: Tool 03 used without being defined at line 1877 in file "/home/jaseg/proj/gerbolyze/gerbonara/gerbonara/tests/resources/allegro-2/MinnowMax_RevA1_DRILL/MinnowMax_RevA1_NCDRILL.drl"
** (process:857403): WARNING **: 11:35:53.851: Setting a default size of 0.04"
** (process:857403): CRITICAL **: 11:35:53.851: Tool 04 used without being defined at line 1924 in file "/home/jaseg/proj/gerbolyze/gerbonara/gerbonara/tests/resources/allegro-2/MinnowMax_RevA1_DRILL/MinnowMax_RevA1_NCDRILL.drl"
** (process:857403): WARNING **: 11:35:53.851: Setting a default size of 0.048"
```

The syntax for these tool definitions is reasonably simple to parse, see [gerbonara's parser](https://gitlab.com/gerbolyze/gerbonara/-/blob/main/gerbonara/excellon.py#L578).

rampageservices:
Allegro EDA exports Excellon drill files with tool definitions embedded
in comment lines rather than standard `T01C0.024` syntax:

```
;T01 Holesize 1. = 8.000000 Tolerance = +3.000000/-3.000000 PLATED MILS Quantity = 1873
```

gerbv previously skipped all comment lines, so these tools were never
registered. When the data section referenced T01, T02, etc., gerbv fell
back to bogus default sizes.

Adds `drill_parse_allegro_comment_tooldef()` which parses these comment
lines during the header section, extracting the tool number and hole
size. Supports MILS, MM, and INCH units (defaults to MILS). Standard
`T01C...` definitions take precedence if both are present.

Fixes #96


LemonBoy:
I have a `.drl` file that I cannot publicly share and that highlights two interesting problems:
- The logic in `drill_file_p` is too strict as the file has no `T` command. 
- The tool definition in the comment can somehow avoid specifying the tool index.

Here's a reduced file that reproduces the problem:
```
;LEADER: 12 
;HEADER: 
;CODE  : ASCII 
;   Holesize 1. = 6.000000 Tolerance = +3.000000/-3.000000 PLATED MILS Quantity = 389
%
G90
X0005906Y0005906
X0019300Y0012100
M30
```

In general such strange files are produced by Allegro and are usually structured as follows:
- Header with tool definitions, with each tool taking a slot
- Usual program without any tool selection command
- `M00` to stop and let the operator change the drill bit by hand
- Other program piece
- Another `M00`
- Lather, rinse and repeat till' `M30` is met

rampageservices:
Thanks for the detailed report @LemonBoy — I've updated the PR to handle all three issues you described.

**Changes:**

1. **`drill_file_p` relaxed** — Files with `Holesize` comments (without T prefix) are now accepted as valid drill files even without any `T` command. Added a `found_holesize` flag that serves as an alternative to `found_T` in the detection logic.

2. **No-T-prefix Holesize comments** — `drill_parse_allegro_comment_tooldef` now handles two formats:
   - `;T01 Holesize 1. = ...` (original, with T prefix)
   - `;   Holesize 1. = ...` (Allegro style, no T prefix — uses the slot index as the tool number)
   
   When the no-T format is detected, a `allegro_noT` flag is set on the parser state to enable M00 tool-change behavior.

3. **M00 as tool change** — In `allegro_noT` mode, `M00` advances to the next defined tool slot instead of ending the program. This handles the Allegro pattern of sequential tool changes without explicit T selection commands.

4. **Auto-select first tool** — When the first X/Y coordinate is encountered in `allegro_noT` mode with no tool selected, the parser scans for the first defined tool and selects it automatically.

5. **No-M48 header support** — The comment tool definition parser now accepts comments in `DRILL_NONE` state (not just `DRILL_HEADER`), since these Allegro files have no `M48` header block.

I've added a test file based on your reduced example (`test-drill-allegro-no-tcommand.exc`) with two tool sizes (24 mil / 60 mil), four holes for tool 1, and two holes for tool 2 separated by M00. All existing drill tests still pass.

spe-ciellt:
Some review comments before merge.

`drill_stats_add_to_drill_list` always passes the inch-converted size with label "inch", so for MM-unit Allegro files the statistics table shows the wrong unit — should use the same unit-conditional logic as drill_parse_T_code() does at every other call site.

Thanks!

PR #337, opened 2026-03-02
fix: Put version number into libgerbv.pc correctly.

eyal0:
libgerbv.pc is used by the `pkg-config` tool to get information about the installed version of libgerbv.  It should look something like this:

```
prefix=/usr/local
exec_prefix=${prefix}
libdir=${prefix}/lib
includedir=${prefix}/include
pkgincludedir=${prefix}/include/gerbv

Name: libgerbv
Description: Core library for gerbv
Requires: glib-2.0 gtk+-2.0
Version: 2.11.0
Libs: -L${libdir} -lgerbv
Cflags: -I${pkgincludedir}
```

The `pkg-config` tool can then extract the version of libgerbv by scanning through all the `*.pc` files on the computer.  It works like this:

```
└──> pkg-config libgerbv --modversion
2.11.0
```

However, if the `Version` in that file is missing, it will report nothing.

The autotools programs use a `configure` script in order to configure a make file for building.  (autotools is an older alternative to cmake.)  One of the commands in the configure script might be like this:

```
PKG_CHECK_MODULES([gerbv], [libgerbv >= 2.1.0])
```

To check that the libgerbv version is at least 2.1.0.

However, if the libgerbv.pc file does not have a version, then it will fail and the user will be unable to complete configuration necessary for compilation.

By fixing libgerbv.pc, pkg-config will be able to report the correct version, and then configure will find that version and be able to create a makefile for programs that use libgerbv as a library.

### DIFF
--- a/src/libgerbv.pc.in
+++ b/src/libgerbv.pc.in
@@ -7,6 +7,6 @@ pkgincludedir=@includedir@/@PACKAGE@
 Name: libgerbv
 Description: Core library for gerbv
 Requires: glib-2.0 gtk+-2.0
-Version: @VERSION@
+Version: @PROJECT_VERSION@
 Libs: -L${libdir} -lgerbv
 Cflags: -I${pkgincludedir}


eyal0:
This has been a *major* issue for users of pcb2gcode.  It causes `configure` tool from automake to fail to find gerbv.

I hope that you can push a new release with this.

eyal0:
https://github.com/pcb2gcode/pcb2gcode/issues/626

spe-ciellt:
Even if you feeling stressed or anything I would be glad if you could fill out the description properly from the beginning. Especially on the Why part.

But I will fix this now.

eyal0:
> Even if you feeling stressed or anything I would be glad if you could fill out the description properly from the beginning. Especially on the Why part.
> 
> But I will fix this now.

No problem!  https://github.com/gerbv/gerbv/pull/337#issue-4012913926

Let me know if that is sufficient.

Issue #334, opened 2026-03-02
Simple fixes for DXF-lib export update


spe-ciellt:
Minor comments that can be solved by another PR:
* It would be great if the patches you applied are documented in the `thirdparty/dxflib/VERSION` file or in a separate `thirdparty/dxflib/PATCHES.md` file.
* Redundant guard in tessellation 
```
  int steps = (int)(fabs(angle2 - angle1) / 2.0) + 1;
  if (steps < 1)
      steps = 1;
```
  The + 1 already guarantees steps >= 1, so the if is dead code. Harmless, but worth removing.

Else great idea with external thirdparty directory. I will merge this.

_Originally posted by @spe-ciellt in https://github.com/gerbv/gerbv/issues/296#issuecomment-3985451048_

rampageservices:
Both items addressed in #340:
- Patches documented in `thirdparty/dxflib/VERSION`
- Dead `if (steps < 1)` guard removed

Regarding upstreaming the patches to dxflib itself — the QCAD GitHub repo doesn't appear to have an issue tracker, so I'm not sure how to report them upstream at this point. Haven't looked further into it yet.

rampageservices:
Separately — tinyscheme (currently vendored directly in `src/`) should probably also be moved into `thirdparty/` for consistency with the dxflib bundling approach. That's a larger refactor though, likely its own issue.

spe-ciellt:
Fix merged and a new issue regarding Tinyscheme created. Should be closed unless I missed something.

PR #340, opened 2026-03-03
fix: document dxflib patches and remove dead tessellation guard

rampageservices:
Addresses #334.

- Documents the two patches applied to the bundled dxflib in `thirdparty/dxflib/VERSION`
- Removes the redundant `if (steps < 1) steps = 1` guard in DXF polygon arc tessellation — the `+ 1` already guarantees `steps >= 1`


spe-ciellt:
Perfect!

Issue #335, opened 2026-03-02
Small fixes on Excellon G02/G03 arc routing with I/J center offsets


spe-ciellt:
Comments for future updates

### 1. Code duplication — `calc_cirseg_mq` and `calc_cirseg_bbox` re-implemented verbatim

Both are `static` in `gerber.c` so they can't be called from `drill.c`. The PR duplicates:
- Full body of `calc_cirseg_mq()` (~30 lines of math + comments)
- Most of `calc_cirseg_bbox()` (~25 lines) reimplemented inline in the bbox block

If a bug is ever found in `calc_cirseg_mq`, it needs fixing in two places. The correct structural fix is to move both functions to a shared location (e.g., `gerb_image.c`) and export them. That's a larger refactor — acceptable to defer, but worth a TODO comment or a follow-up issue.

### 2. Bounding box uses `floor()` — differs from `calc_cirseg_bbox` but is actually more correct

The PR's middle-point loop:
```c
for (step_pi_2 = (floor(ang1 / M_PI_2) + 1) * M_PI_2; ...)
```
`calc_cirseg_bbox` in `gerber.c` uses:
```c
for (step_pi_2 = (ang1/M_PI_2 + 1)*M_PI_2; ...)
```
For non-axis-aligned `ang1` (e.g. 45°), `gerber.c` gives start = `1.5×(π/2)` = 135° — which **skips the 90° axis crossing**. The PR's `floor(0.5)=0` gives start = `1×(π/2)` = 90° — which is **correct** (first axis crossing after 45°). The PR accidentally fixes a latent imprecision in the original bbox algorithm. No action needed here, but it's worth documenting.

### 3. `delta_cp_x/y` not reset on T-code tool change (`drill.c:887`)

```c
case 'T':
    drill_parse_T_code(fd, state, image, file_line);
    state->route_mode = DRILL_G_DRILL;
    state->tool_down = FALSE;
    /* delta_cp_x/y NOT reset here */
    break;
```
In practice this is harmless — both fields are cleared after every arc segment and on tool-up, so by the time a T-code is seen they're already 0. But for defensive consistency with the `route_mode`/`tool_down` resets, adding `state->delta_cp_x = state->delta_cp_y = 0;` here would be better hygiene.

Great patch, will merge it and drop an issue for future reference.

_Originally posted by @spe-ciellt in https://github.com/gerbv/gerbv/issues/298#issuecomment-3985595432_

rampageservices:
[...]

PR #342, opened 2026-03-03
fix: Correct computation for bounding box of gerber files.

### LINKED ISSUE #336: Incorrect gerber bounds computed by gerbv
For this file, gerbv is computing the wrong bounds:

```
%FSLAX46Y46*%
%MOMM*%
%LPD*%
G01*
%ADD10C,0.100000*%
D10*
X-20000000Y-20000000D02*
G75*
G02*
Y20000000I20000000J20000000D01*
X20000000I20000000J-20000000D01*
Y-20000000I-20000000J-20000000D01*
X-20000000I-20000000J20000000D01*
M02*
```

That gerber file is a drawing of a circle made by four arcs.  But the arcs are on in the quadrants, they are arcs at north, south, east and west.  The cricle circumscribes an aligned square with opposite corners are 2,2 and -2,-2.

Image

In my testing, gerbv is reporting that the bounding box for the shape is too small:

Bounding box: POLYGON((-0.86811 -0.86811,-0.86811 0.86811,0.86811 0.86811,0.86811 -0.86811,-0.86811 -0.86811))

The issue was found here: https://github.com/pcb2gcode/pcb2gcode/issues/784

I will look at it

eyal0:
For the bounding box of an arc, we must consider the starting point, the end point, and every point that is at an x-axis or y-axis extent. Those occur at the angles that are multiples of 90 degrees.  The `for` loop should iterate over those rather than every 90 degrees start from the start of the arc.

Similar code is in `drill.c`.

This fixes #336.

### DIFF
--- a/src/gerber.c
+++ b/src/gerber.c
@@ -2538,7 +2538,7 @@ calc_cirseg_mq(struct gerbv_net *net, int cw,
     /*
      * This is a sanity check for angles after the nature of atan2.
      * If cw we must make sure angle1-angle2 are always positive,
-     * If ccw we must make sure angle2-angle1 are always negative.
+     * If ccw we must make sure angle2-angle1 are always positive.
      * We should really return one angle and the difference as GTK
      * uses them. But what the heck, it works for me.
      */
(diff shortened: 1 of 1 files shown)


rampageservices (approved):
LGTM. This matches the corrected version already in `drill.c:502` (from #298), and fixes the exact bug we described in #336.

For reference, #335 point 2 also documented this discrepancy between the two implementations — this PR resolves it.

PR #356, opened 2026-03-05
fix: suppress GdkPixbuf XPM loader warning on Windows

rampageservices:
## Summary

Suppresses the "GdkPixbuf-WARNING: Error loading XPM image loader" warning on Windows startup.

gerbv doesn't use any GdkPixbuf loaders — image export uses Cairo directly, and the only XPM usage is inline data parsed by `gdk_pixmap_create_from_xpm_d()`. The warning is harmless but confusing to users.

### Approach (reworked)

The original PR shipped an empty `loaders.cache` file via the packaging scripts, but this broke when the Windows build switched from ZIP to NSIS installer (#441, #462).

The reworked fix is entirely in code — at startup (before `gtk_init`), gerbv:
1. Creates `%LOCALAPPDATA%/gerbv/loaders.cache` with an empty GdkPixbuf module list
2. Sets `GDK_PIXBUF_MODULE_FILE` to point at it (only if the env var isn't already set)

This tells GdkPixbuf there are no loaders to enumerate, suppressing the warning. No packaging changes needed — works with both the NSIS installer and the MSYS2 ZIP distribution.

Single file change: `src/main.c`, guarded by `#ifdef WIN32`.

## Test plan

- [x] Builds clean with `-Werror` on Linux
- [x] No new test regressions
- [ ] Verify on Windows: XPM loader warning no longer appears at startup
- [ ] Verify `GDK_PIXBUF_MODULE_FILE` override still works (user-set value is respected)

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -603,7 +603,33 @@ main(int argc, char *argv[])
 		    NULL);
     g_setenv(settings_schema_env, env_val, TRUE);
     g_free(env_val);
-    
+
+#ifdef WIN32
+    /* Suppress "GdkPixbuf-WARNING: Error loading XPM image loader"
+     * on Windows.  gerbv doesn't use any GdkPixbuf loaders — image
+     * export uses Cairo, and the only XPM usage is inline data via
+     * gdk_pixmap_create_from_xpm_d().  Point GDK_PIXBUF_MODULE_FILE
+     * at an empty cache so GdkPixbuf skips loader enumeration. */
+    if (!g_getenv("GDK_PIXBUF_MODULE_FILE")) {
+	gchar *cache_dir = g_build_filename(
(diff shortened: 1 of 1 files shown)


rampageservices:
Reworked this PR from scratch. The old approach (shipping `loaders.cache` via packaging scripts) broke when the Windows build switched from ZIP to NSIS installer (#441, #462).

The new approach is entirely in code — at startup, gerbv creates an empty `loaders.cache` in the user cache dir and sets `GDK_PIXBUF_MODULE_FILE` before `gtk_init`. No packaging changes needed, works with both NSIS and MSYS2 distributions.

Force-pushed: old 2-file packaging diff → single-file `main.c` change (27 lines).

PR #365, opened 2026-03-05
fix: Build authors.c before all locales.

eyal0:
Parallel build is failing without this because the `*.pot` files require `config/authors.c`.

### DIFF
--- a/locale/CMakeLists.txt
+++ b/locale/CMakeLists.txt
@@ -67,6 +67,9 @@ configure_gettext(
  ALL
 )
 add_dependencies(${GETTEXT_TARGET} generated)
+foreach(lang IN LISTS GETTEXT_LANGUAGES)
+    add_dependencies(${GETTEXT_TARGET}-${lang} generated)
+endforeach()
 
 install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/${GETTEXT_OUTPUT_DIR}/ TYPE LOCALE)
 


eyal0:
To reproduce this bug, from the root of gerbv, type:

```
git clean -i -x -d
mkdir build
cd build
cmake ..
make -j 20
```

The output of the `make` will be:

```
└──> make -j 20
make[2]: *** No rule to make target 'config/authors.c', needed by '/home/esoha/private/gerbv/locale/locale/gerbv.pot'.  Stop.
make[1]: *** [CMakeFiles/Makefile2:364: locale/CMakeFiles/gerbv-translations-ru.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
make[2]: *** No rule to make target 'config/authors.c', needed by '/home/esoha/private/gerbv/locale/locale/gerbv.pot'.  Stop.
make[1]: *** [CMakeFiles/Makefile2:332: locale/CMakeFiles/gerbv-translations-ja.dir/all] Error 2
[  2%] Building CXX object CMakeFiles/dxflib_bundled.dir/thirdparty/dxflib/dl_dxf.cpp.o
[  4%] Generating bugs.c, authors.c
[  6%] Building CXX object CMakeFiles/dxflib_bundled.dir/thirdparty/dxflib/dl_writer_ascii.cpp.o
[  9%] Building C object CMakeFiles/tinyscheme_bundled.dir/thirdparty/tinyscheme/dynload.c.o
[ 11%] Building C object CMakeFiles/tinyscheme_bundled.dir/thirdparty/tinyscheme/scheme.c.o
AUTHORS GENERATED /home/esoha/private/gerbv/AUTHORS => /home/esoha/private/gerbv/build/config/authors.c
BUGS GENERATED /home/esoha/private/gerbv/BUGS => /home/esoha/private/gerbv/build/config/bugs.c
[ 11%] Built target generated
/home/esoha/private/gerbv/thirdparty/tinyscheme/scheme.c: In function ‘Cisalpha’:
/home/esoha/private/gerbv/thirdparty/tinyscheme/scheme.c:269:44: warning: implicit declaration of function ‘isascii’ [-Wimplicit-function-declaration]
  269 | static INLINE int Cisalpha(int c) { return isascii(c) && isalpha(c); }
      |                                            ^~~~~~~
[ 13%] Linking C static library libtinyscheme_bundled.a
[ 13%] Built target tinyscheme_bundled
[ 15%] Linking CXX static library libdxflib_bundled.a
[ 15%] Built target dxflib_bundled
make: *** [Makefile:166: all] Error 2
Exit code: 2                                                                                                                                                                                                                           
```

We can see that it is trying to build gerbv.pot before the authors.c has been created.  This is a mistake.

rampageservices (approved):
LGTM. The race is between the per-language targets and the `generated` custom target — `add_dependencies(gerbv-translations generated)` only orders `generated` before `gerbv-translations` itself, not before its children `gerbv-translations-ja`/`gerbv-translations-ru`, so in a parallel build `xgettext` can fire before `authors.c` exists.

Adding the dependency to each per-language target is the correct fix.

Nit (not blocking): the existing `add_dependencies(${GETTEXT_TARGET} generated)` on line 69 is now redundant since all children depend on `generated` and the parent depends on all children. Could be removed for clarity, but harmless either way.

eyal0:
> the existing `add_dependencies(${GETTEXT_TARGET} generated)` on line 69 is now redundant since all children depend on `generated` and the parent depends on all children. Could be removed for clarity, but harmless either way.

cmake always confuses me so I just had the AI code up the fix.  Seems to work!

I'm finding these errors because I use libgerbv and not gerbv.  Also, I am compiling for CI on Windows under msys2.  These are probably uncommon platforms!  So I'm getting these uncommon errors.

PR #367, opened 2026-03-05
fix: Include dxf bundle in libgerbv.

eyal0:
If dxf is not found on the system, include it in libgerbv.a.

This fixes #366.

PR #368, opened 2026-03-06
fix: Include header files to fix compile errors.

eyal0:
The windows compilation seems more sensitive to missing header files than linux compilation.


eyal0:
This is a draft until the underlying PR #367 is in.

rampageservices (commented):
The `math.h`/`_USE_MATH_DEFINES` fix (commit 2) looks good.

A few notes on the dxflib bundling (commit 1):

**`COMPILE_WARNING_AS_ERROR OFF` is a no-op.** The project doesn't set `-Werror` or `COMPILE_WARNING_AS_ERROR` anywhere, so there's nothing to turn off. This can be dropped.

**`PUBLIC` include dir on the OBJECT library is dead.** OBJECT library usage requirements don't propagate through `$<TARGET_OBJECTS:>`, which is why you had to add the explicit `target_include_directories(gerbv PRIVATE ...)` in `src/CMakeLists.txt`. The `PUBLIC` on `dxflib_bundled` no longer does anything.

**Inconsistent path style.** The root `CMakeLists.txt` changed `${CMAKE_SOURCE_DIR}/thirdparty` to the relative `thirdparty`, but `src/CMakeLists.txt` uses `${CMAKE_SOURCE_DIR}/thirdparty`. Both work, but the inconsistency is unnecessary.

**Simpler alternative: skip the OBJECT library entirely.** Since the only goal is embedding dxflib symbols into `libgerbv.a`, you can just add the sources directly to the gerbv target:

```cmake
# Root CMakeLists.txt — replace the add_library block:
if(NOT DXF_FOUND)
    set(DXFLIB_BUNDLED_SOURCES
        ${CMAKE_SOURCE_DIR}/thirdparty/dxflib/dl_dxf.cpp
        ${CMAKE_SOURCE_DIR}/thirdparty/dxflib/dl_writer_ascii.cpp)
    set(DXF_FOUND TRUE)
    set(DXFLIB_BUNDLED TRUE)
endif()

# src/CMakeLists.txt — in the DXF_FOUND block:
if(DXFLIB_BUNDLED)
    target_sources(gerbv PRIVATE ${DXFLIB_BUNDLED_SOURCES})
    target_include_directories(gerbv PRIVATE ${CMAKE_SOURCE_DIR}/thirdparty)
endif()
```

This eliminates the `dxflib_bundled` target, the OBJECT library, the dead `PUBLIC` include propagation, and the unnecessary `COMPILE_WARNING_AS_ERROR OFF` — while achieving the same result.

rampageservices (commented):
The `math.h`/`_USE_MATH_DEFINES` fix (commit 2) looks good.

A few notes on the dxflib bundling (commit 1):

**`COMPILE_WARNING_AS_ERROR OFF` is a no-op.** The project doesn't set `-Werror` or `COMPILE_WARNING_AS_ERROR` anywhere, so there's nothing to turn off. This can be dropped.

**`PUBLIC` include dir on the OBJECT library is dead.** OBJECT library usage requirements don't propagate through `$<TARGET_OBJECTS:>`, which is why you had to add the explicit `target_include_directories(gerbv PRIVATE ...)` in `src/CMakeLists.txt`. The `PUBLIC` on `dxflib_bundled` no longer does anything.

**Inconsistent path style.** The root `CMakeLists.txt` changed `${CMAKE_SOURCE_DIR}/thirdparty` to the relative `thirdparty`, but `src/CMakeLists.txt` uses `${CMAKE_SOURCE_DIR}/thirdparty`. Both work, but the inconsistency is unnecessary.

**Simpler alternative: skip the OBJECT library entirely.** Since the only goal is embedding dxflib symbols into `libgerbv.a`, you can just add the sources directly to the gerbv target:

```cmake
# Root CMakeLists.txt — replace the add_library block:
if(NOT DXF_FOUND)
    set(DXFLIB_BUNDLED_SOURCES
        ${CMAKE_SOURCE_DIR}/thirdparty/dxflib/dl_dxf.cpp
        ${CMAKE_SOURCE_DIR}/thirdparty/dxflib/dl_writer_ascii.cpp)
    set(DXF_FOUND TRUE)
    set(DXFLIB_BUNDLED TRUE)
endif()

# src/CMakeLists.txt — in the DXF_FOUND block:
if(DXFLIB_BUNDLED)
    target_sources(gerbv PRIVATE ${DXFLIB_BUNDLED_SOURCES})
    target_include_directories(gerbv PRIVATE ${CMAKE_SOURCE_DIR}/thirdparty)
endif()
```

This eliminates the `dxflib_bundled` target, the OBJECT library, the dead `PUBLIC` include propagation, and the unnecessary `COMPILE_WARNING_AS_ERROR OFF` — while achieving the same result.

rampageservices:
Thanks for identifying the root cause here @eyal0 — I've put up #369 as a simplified alternative that compiles the bundled dxflib sources directly into the `gerbv` library target via `target_sources()` instead of using an intermediate OBJECT library. This avoids the extra CMake target while achieving the same result (symbols embedded in `libgerbv.a`).

I've credited you as co-author on both commits.

eyal0:
> COMPILE_WARNING_AS_ERROR OFF is a no-op. The project doesn't set -Werror or COMPILE_WARNING_AS_ERROR anywhere, so there's nothing to turn off. This can be dropped.

I think that this is not true.

ci.yml uses a preset for cmake: https://github.com/gerbv/gerbv/blob/develop/.github/workflows/ci.yaml#L65

That preset is here: https://github.com/gerbv/gerbv/blob/develop/CMakePresets.json#L4

It loads this: https://github.com/gerbv/gerbv/blob/develop/cmake/preset/linux-gnu-gcc.json#L3

Which loads the base.json: https://github.com/gerbv/gerbv/blob/develop/cmake/preset/base.json#L10

eyal0:
I do prefer your simple alternative.  Go for it!

rampageservices:
You were right about `COMPILE_WARNING_AS_ERROR` — the preset chain sets it in `base.json`. I missed that during the initial review, apologies.

Updated #369 to handle it: the bundled dxflib sources now get `-Wno-uninitialized` / `-Wno-maybe-uninitialized` via `set_source_files_properties` so they don't trip `-Werror` while keeping it enforced for gerbv's own code.

Also moved the `_USE_MATH_DEFINES` fix from `gerbv.h` to a CMake-level compile definition on MSVC — the header approach doesn't work because `draw.c`, `draw-gdk.c`, and `export-isel-drill.c` include `<math.h>` before `gerbv.h`, so the define arrives too late.

rampageservices:
Heads up — spe-ciellt opened #374 to restructure all thirdparty library handling into `thirdparty/*/CMakeLists.txt` (tracking issue #373). He'd prefer that restructuring lands first, with other CMake PRs rebasing on top.

I've closed #369 in favor of that approach and left review comments on #374 noting the fixes that still need to go in (dxflib embedding, warning suppression, `_GNU_SOURCE`, MSVC `_USE_MATH_DEFINES`).

You may want to close this one too and coordinate with #374 directly, since the OBJECT library approach here would conflict with the new directory layout.

eyal0:
Leaving this here for reference: This branch of gerbv successfully compiles on both ubuntu and mingw32 windows.  Also, the libgerbv.a created on both platforms works successfully in pcb2gcode.

The necessary changes are:

* Compile the dxflib bundle directly into the code, not as a library (https://github.com/gerbv/gerbv/pull/368#pullrequestreview-3901118193)
* Make sure that, if cmake is running on `WIN32`, set the variable `WIN32`: https://github.com/gerbv/gerbv/pull/368/changes#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR77-R79
* Define `M_PI`, `M_PI_2`, and `M_1_PI` in gerbv.h https://github.com/gerbv/gerbv/pull/368/changes#diff-3802cbcdb0af8c0a04ec0fc67cab4d18dbed5f686c2b6cad1d836323c0b64bbdR81-R90

It might be possible to get M_PI et al in another way but `_USE_MATH_DEFINES` didn't work for me.  Maybe WIN32 is sufficient?  I didn't try that.  `G_PI` and `G_PI_2` exist but I think that it is less accurate: https://docs.gtk.org/glib/const.PI.html .  Mine are copy-pasted from math.h.

rampageservices:
I reviewed the current state of this PR. After the M_PI fix was extracted into #393 (merged), the diff is down to 2 files / 10 lines:

### What remains

**1. `CMakeLists.txt`** — platform defines on the `compilation-flags` interface target:
```cmake
if(APPLE)
    target_compile_definitions(compilation-flags INTERFACE QUARTZ)
endif()
if(WIN32)
    target_compile_definitions(compilation-flags INTERFACE WIN32)
endif()
    target_compile_definitions(compilation-flags INTERFACE _GNU_SOURCE)
endif()
```

**2. `thirdparty/tinyscheme/scheme.c`** — moves `#include <stdlib.h>` before `#include <stddef.h>` so `_CRTIMP` is defined on MinGW.

### Concerns

**`WIN32`/`QUARTZ` was explicitly rejected by the maintainer.** @spe-ciellt closed #414 (which had the same change) saying platform/toolchain specifics don't belong in top-level `CMakeLists.txt` — they belong in toolchain files. Keeping them here will get the same pushback.

**`_GNU_SOURCE` can be merged** — spe-ciellt said as much in the #414 discussion. But it's bundled with the rejected changes here.

**Commit history** — 2 "WIP" messages, 3 merge commits. The tinyscheme include reorder is a real MinGW fix but has no explanatory commit message.

**PR description is stale** — "The windows compilation seems more sensitive to missing header files" doesn't describe what the PR does now.

### Suggestion

Consider splitting the remaining work:
1. **`_GNU_SOURCE` only** — fresh single-commit PR, which the maintainer has indicated he'll accept
2. **tinyscheme `stdlib.h` order** — separate commit with a message explaining the MinGW `_CRTIMP` dependency
3. **Drop `WIN32`/`QUARTZ`** from CMakeLists.txt — these were rejected in #414; the maintainer wants them in toolchain files

That would unblock the non-controversial parts instead of keeping everything stuck behind the `WIN32`/`QUARTZ` disagreement.

PR #369, opened 2026-03-06
fix: embed bundled dxflib directly into libgerbv

### LINKED ISSUE #366: libgerbv is not installing correctly
As a program, `gerbv` works.  As a library, though, it isn't working.  I'm investigating it.

It is used by `pcb2gcode`.

rampageservices:
## Summary

When system `dxflib` is absent, the bundled copy is built as a separate `STATIC` library (`dxflib_bundled`) and linked via the `compilation-flags` INTERFACE target. The gerbv executable links fine, but `libgerbv.a` never embeds the dxflib object files — they show up as `*UND*`, breaking downstream consumers like pcb2gcode.

This PR:
- Replaces the separate `dxflib_bundled` STATIC library with a `DXFLIB_BUNDLED_SOURCES` variable
- Adds the bundled sources directly to the `gerbv` library target via `target_sources()`, so symbols are embedded in `libgerbv.a`
- Suppresses `-Wno-uninitialized` on the bundled dxflib files (GCC false positive on `DL_HatchEdgeData` default construction, tracked in #372)
- Defines `_USE_MATH_DEFINES` on MSVC via `compilation-flags` so `M_PI`/`M_1_PI` (used by `DEG2RAD`/`RAD2DEG`) are available under strict ANSI conformance — a header-level define in `gerbv.h` doesn't work because `draw.c`, `draw-gdk.c`, and `export-isel-drill.c` include `<math.h>` before `gerbv.h`

Fixes #366

## Approach

@eyal0's PR #368 identified the same root cause and proposed an `OBJECT` library approach. This PR takes a simpler path: compiling the bundled `.cpp` files directly into `libgerbv` via `target_sources()`, avoiding the extra CMake target entirely.

## Test plan

- [ ] `cmake --build` with no system dxflib → bundled sources compile into libgerbv
- [ ] `nm libgerbv.a | grep dl_dxf` → symbols are defined (`T`), not undefined (`U`)
- [ ] `cmake --build` with system dxflib installed → pkg-config path still works
- [ ] CI passes with `CMAKE_COMPILE_WARNING_AS_ERROR ON` (no dxflib warnings)
- [ ] MSVC build → `M_PI` resolves via the compile definition

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,13 +23,11 @@ pkg_check_modules(GLIB REQUIRED IMPORTED_TARGET glib-2.0>=2.6)
 ## DXF support library
 pkg_check_modules(DXF IMPORTED_TARGET dxflib)
 if(NOT DXF_FOUND)
-    # Bundled dxflib (GPL-2+, RibbonSoft/QCAD)
-    add_library(dxflib_bundled STATIC
-        thirdparty/dxflib/dl_dxf.cpp
-        thirdparty/dxflib/dl_writer_ascii.cpp)
-    set_target_properties(dxflib_bundled PROPERTIES POSITION_INDEPENDENT_CODE ON)
-    target_include_directories(dxflib_bundled PUBLIC
-        ${CMAKE_SOURCE_DIR}/thirdparty)
+    # Bundled dxflib sources (GPL-2+, RibbonSoft/QCAD) — compiled directly
+    # into libgerbv so symbols are embedded in the static archive.
+    set(DXFLIB_BUNDLED_SOURCES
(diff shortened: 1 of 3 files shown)


eyal0:
That math fix isn't working for me yet, don't merge it yet please.

rampageservices:
Closing in favor of #374 — spe-ciellt is restructuring the thirdparty CMake layout (#373) and it makes sense to land that first.

The bug fixes from this PR (dxflib symbol embedding, warning suppression, MSVC `_USE_MATH_DEFINES`) are noted on #374 for incorporation. The `target_sources()` approach for dxflib embedding still applies — #374's current STATIC library won't fix the undefined symbols in `libgerbv.a`.

PR #371, opened 2026-03-06
fix: define _GNU_SOURCE for tinyscheme on Linux

### LINKED ISSUE #370: tinyscheme fails to compile with GCC 15 / C23: isascii undeclared
## Problem

Building gerbv with GCC 15 on Linux fails with:

```
thirdparty/tinyscheme/scheme.c: In function 'Cisalpha':
thirdparty/tinyscheme/scheme.c:269:44: error: implicit declaration of function 'isascii' [-Wimplicit-function-declaration]
  269 | static INLINE int Cisalpha(int c) { return isascii(c) && isalpha(c); }
      |                                            ^~~~~~~
```

`isascii` is a POSIX function, not part of the C standard. The project sets `CMAKE_C_STANDARD 23` with `CMAKE_C_EXTENSIONS OFF`, so GCC compiles with `-std=c23` which doesn't expose POSIX extensions. In C23, implicit function declarations are a hard error.

## Root cause

The `compilation-flags` INTERFACE target defines `_GNU_SOURCE` on Linux (which exposes `isascii` from `<ctype.h>`), but `tinyscheme_bundled` links only `config` — it never gets `_GNU_SOURCE`.

## Fix

Define `_GNU_SOURCE` on the `tinyscheme_bundled` target for Linux, matching the existing pattern in `compilation-flags`.

rampageservices:
## Summary

Building with GCC 15 on Linux fails because `tinyscheme_bundled` calls `isascii()`, a POSIX function not exposed under `-std=c23` (no extensions). GCC 15 treats implicit function declarations as a hard error in C23.

The `compilation-flags` INTERFACE target already defines `_GNU_SOURCE` on Linux, but `tinyscheme_bundled` links only `config` and never receives it. This adds the definition directly to the tinyscheme target.

Fixes #370

## Test plan

- [ ] `cmake --build` with GCC 15 on Linux — tinyscheme compiles without `-D_GNU_SOURCE` workaround
- [ ] Verify no regression on GCC 14 or Clang

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,6 +48,9 @@ target_include_directories(tinyscheme_bundled PUBLIC
 target_include_directories(tinyscheme_bundled PRIVATE
     ${CMAKE_SOURCE_DIR}/src)
 target_link_libraries(tinyscheme_bundled PRIVATE config)
+if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
+    target_compile_definitions(tinyscheme_bundled PRIVATE _GNU_SOURCE)
+endif()
 if(Intl_FOUND)
     target_link_libraries(tinyscheme_bundled PRIVATE Intl::Intl)
 endif()


rampageservices:
Closing in favor of #374 — the `_GNU_SOURCE` fix for tinyscheme should go in the new `thirdparty/tinyscheme/CMakeLists.txt` that #374 introduces. Noted on that PR.

Issue #372, opened 2026-03-06
Bundled dxflib: DL_HatchEdgeData triggers -Wuninitialized with -Werror


rampageservices:
## Problem

The bundled dxflib code emits `-Wuninitialized` / `-Wmaybe-uninitialized` warnings when compiled with `-Wall`:

```
dl_entities.h:1539:22: error: '<anonymous>.DL_HatchEdgeData::type' is used uninitialized [-Werror=uninitialized]
 1539 | struct DXFLIB_EXPORT DL_HatchEdgeData {
```

Triggered at `dl_dxf.cpp:606` and `dl_dxf.cpp:2018` where `hatchEdge = DL_HatchEdgeData()` default-constructs the struct. GCC flags the implicit move-assignment of uninitialized members as a potential use-before-init, even though the struct is being reset (not read).

This is a false positive in GCC's analysis — the default constructor value-initializes all members — but it fires reliably with `-O3` inlining.

## Current workaround

PR #369 suppresses `-Wno-uninitialized` and `-Wno-maybe-uninitialized` on the bundled dxflib source files via `set_source_files_properties`. This keeps `-Werror` working for gerbv's own code.

## Upstream fix

The proper fix is in dxflib itself (QCAD): either adding a user-defined default constructor to `DL_HatchEdgeData` that explicitly initializes all members, or using `= {}` member initializers. This should be reported to https://github.com/qcad/qcad.

eyal0:
Good plan!

PR #374, opened 2026-03-06
CMake for thirdparty libraries

### LINKED ISSUE #373: Reorganize CMake for third party libraries first
There is a lof of handling of the third party libraries in the top level `CMakeLists.txt`. 

The top level `CMakeLists.txt` should optimally only have  a `add_subdirectory(thirdparty)`. In the `thirdparty` directory there should be a `CMakeLists.txt` which contains
```
add_subdirectory(tinyscheme)
add_subdirectory(dxflib)
```
Then respective thirdparty directories should have a `CMakeLists.txt` with all settings for that library in there. And be careful about `PRIVATE` and `PUBLIC` on what is exported. All code in the toplevel `CMakeLists.txt` should move to respective libraries `CMakeLists.txt`.

Both thirdparty libraries should then create a static library like `add_library(<library name> STATIC)` and then be linked in every place it is needed.

I also question if we need to check for (and use) other DXF libs if we have it bundled into the code. Any suggestions or reasons to keep looking for it?

Maybe I'm rude, but I would also suggest to do these changes first, then rebase all PR's involving CMake work for the thirdparty libraries onto the new structure.

spe-ciellt:
Update for CMake  which is more best practices on how to include  third party packages in `CMake`, see my description in #373 .

This will be set in Draft mode until it has been tested and cleaned up a bit. There are probably naming that is broken, commented out code, and pure errors/bugs/omissions in this.

Comments and suggestions are welcome.

Closes #373


rampageservices:
The directory restructuring looks good — centralizing thirdparty build logic into `thirdparty/*/CMakeLists.txt` is cleaner than having it all in the root CMakeLists.txt.

A few things this PR will need before it can close #366:

**dxflib symbol embedding (#366)**: `add_library(dxflib_bundled STATIC)` still produces a separate archive — its symbols won't end up in `libgerbv.a`, which is the original bug. Downstream consumers like pcb2gcode get undefined references. Two options:
- Compile bundled sources directly into the `gerbv` target via `target_sources()` (what #369 did)
- Use an OBJECT library and consume via `target_link_libraries(gerbv PRIVATE dxflib_bundled)` — CMake 3.12+ propagates object files from OBJECT libraries through normal linking

**Warning suppression on bundled dxflib (#372)**: The bundled sources trigger a GCC false positive (`-Wuninitialized` on `DL_HatchEdgeData` default construction). With `CMAKE_COMPILE_WARNING_AS_ERROR ON` in the CI preset, this becomes a build error. Needs `set_source_files_properties` or SYSTEM includes to suppress.

**tinyscheme `_GNU_SOURCE` (#370)**: Building with GCC 15 / `-std=c23` on Linux fails because `isascii()` isn't exposed without `_GNU_SOURCE`. The tinyscheme target needs:
```cmake
    target_compile_definitions(tinyscheme PRIVATE _GNU_SOURCE)
endif()
```

**MSVC `_USE_MATH_DEFINES`**: `M_PI` and `M_1_PI` (used by `DEG2RAD`/`RAD2DEG` in gerbv.h) aren't defined under MSVC strict ANSI. This needs a compile definition on `compilation-flags`:
```cmake
if(MSVC)
    target_compile_definitions(compilation-flags INTERFACE _USE_MATH_DEFINES)
endif()
```

I'm closing #369 and #371 in favor of this restructuring — the fixes above cover what those PRs addressed. Happy to help if needed.

spe-ciellt:
`_GNU_SOURCE`: If you configure with `--preset linux-gnu-gcc`, `_GNU_SOURCE` is included in all compilations. Linux specific settings and Apple specific settings should preferably go into the respective toolchains files. Also valid for MSVC above.

I quickly looked up the `isascii()` and there seems to be more generic ways to fix it.

I have added the `OBJECT` flag now as well.

rampageservices:
Makes sense — replacing `isascii()` with a portable equivalent like `((unsigned)(c) <= 0x7f)` in `scheme.c` would be cleaner than relying on `_GNU_SOURCE`. That patch would go into `GERBV_PATCHES` alongside the existing vendored dxflib patches.

One small thing: `thirdparty/CMakeLists.txt` is missing a trailing newline at EOF — some tools (GCC, Git, various linters) will warn about that.

Everything else looks good after the squash. The OBJECT library for dxflib is the right call. Will watch for the `HAVE_LIBDXFLIB` guard cleanup whenever that lands.

Thanks for pushing this through — the thirdparty restructuring is a big improvement.

Issue #352, opened 2026-03-04
Incorrect display of rotated circle in aperture macro


jubilee4321:
```
$ gerbv --version
gerbv version v2.11.1
Copyright (C) 2001-2008 by Stefan Petersen
and the respective original authors listed in the source files.

$ cat code1_circle.gbr
%MOIN*%
%FSLAX23Y23*%
%OFA0.0000B0.0000*%
G90*
%AMCIRCLE*
1,1,2.59,2,5,19.5*
%
%ADD10C,0.0650*%
%ADD11CIRCLE,.5*%
D11
X12Y9D03*
X0Y0D02*
G54D10*
X0Y0D01*
X10000D01*
Y10000D01*
X0D01*
Y0D01*
M02*
```
The display shows the circle inside the square
<img width="687" height="650" alt="Image" src="https://github.com/user-attachments/assets/9ed8aa46-b993-4079-a164-815323888e56" />
The circle should be on the edge
<img width="608" height="555" alt="Image" src="https://github.com/user-attachments/assets/afb09324-e155-4a1e-86db-202145d0b5c5" />
When using the selection tool the circle can only be selected when the cursor is pointing to where the circle should be, that is, where the circle is displayed is not a selection target for the circle.

rampageservices:
This is a known issue — the center coordinates of circle (and line) primitives in aperture macros are being rotated twice during image transformation.

The rendering code in `draw.c` applies rotation by rotating the Cairo coordinate system first, then translating to the center in the rotated space. Since the translation already happens in rotated coordinates, the center should not be pre-rotated. But `gerb_image.c` applies `gerbv_rotate_coord()` on the center coordinates *in addition to* updating the rotation parameter, which double-rotates them.

Your test case `1,1,2.59,2,5,19.5` (center at 2,5 with 19.5° rotation) hits this exactly — the circle renders at the wrong position because the center gets rotated twice.

This is fixed by #310, which removes the redundant `gerbv_rotate_coord()` calls for CIRCLE and LINE21 primitives.

spe-ciellt:
I tested this after merging in #310, I wanted to use it as a potential test case. But it still showed the top level image. I realized it differs between which renderer you use. So it looks wrong in GDK-rendering (fast), but correct in Normal rendering (Cairo).

rampageservices:
@spe-ciellt Good catch — you're right that #310 only fixed the Cairo renderer.

[...]

Issue #233, opened 2024-01-15
rotate macro_line tiwce when export image to rs274x


chenshiqu:
case GERBV_APTYPE_MACRO_LINE22:
						/* Lower left line rectangle */

					/* Using LINE21 parameters array
					 * indexes for LINE21 and LINE22, as
					 * they are equal */
					if (trans->scaleX == trans->scaleY) {
						sam->parameter[LINE21_WIDTH] *=
								trans->scaleX;
						sam->parameter[LINE21_HEIGHT] *=
								trans->scaleX;

					} else if (fabs(sam->parameter[LINE21_ROTATION]) == 0
					|| fabs(sam->parameter[LINE21_ROTATION]) == 180) {
						sam->parameter[LINE21_WIDTH] *=
								trans->scaleX;
						sam->parameter[LINE21_HEIGHT] *=
								trans->scaleY;

					} else if (fabs(sam->parameter[LINE21_ROTATION]) == 90
					|| fabs(sam->parameter[LINE21_ROTATION]) == 270) {
						double t;
						t = sam->parameter[LINE21_WIDTH];
						sam->parameter[LINE21_WIDTH] =
							trans->scaleY *
							sam->parameter[
								LINE21_HEIGHT];
						sam->parameter[LINE21_HEIGHT] =
							trans->scaleX * t;
					} else {
						/* TODO: make outline macro */
						err_scale_line_macro++;
						break;
					}

					sam->parameter[LINE21_CENTER_X] *=
								trans->scaleX;
					sam->parameter[LINE21_CENTER_Y] *=
								trans->scaleY;

					sam->parameter[LINE21_ROTATION] +=
						RAD2DEG(trans->rotation);
                                        /************************ operation here is redundant  *******************************/
					gerbv_rotate_coord(
					        sam->parameter +LINE21_CENTER_X,
					 	sam->parameter +LINE21_CENTER_Y,
					 	trans->rotation);
                                       /*************************************************************/
					break;

eyal0:
Are you able to test it? Why do you say that it is redundant?

chenshiqu:
you can test this file. Use gerbv to load this file, rotate 180 degree, then export to rx-274x. Reload the exported file, you can see the issue.












At 2024-01-16 10:01:11, "Eyal" ***@***.***> wrote:

Are you able to test it? Why do you say that it is redundant?

eyal0:
Where's the file?

chenshiqu:
see the attachment.










[...]

PR #375, opened 2026-03-06
Fix missing rotation for circle primitives in GDK renderer

rampageservices:
## Summary

- `gerbv_gdk_draw_prim1` (circle) was the only GDK macro primitive handler that did not apply `rotate_point()` to its center coordinates
- Before #310, this was masked because `gerb_image.c` pre-rotated the center via `gerbv_rotate_coord()` — the GDK renderer got already-rotated coordinates
- #310 correctly removed that pre-rotation to fix the Cairo double-rotation bug (#233), but exposed that the GDK renderer never had its own rotation handling for circles
- Apply `rotate_point()` to the center offset before translating, matching the pattern used by all other GDK macro primitives (prim4, prim5, prim6, prim7, prim20, prim21, prim22)

## Test plan

- Load the test case from #352 (`1,1,2.59,2,5,19.5` — circle at center 2,5 with 19.5° rotation)
- Switch between Fast (GDK) and Normal (Cairo) rendering — both should now show the circle in the same position on the edge of the square
- Verify selection target matches the rendered position

### DIFF
--- a/src/draw-gdk.c
+++ b/src/draw-gdk.c
@@ -91,17 +91,22 @@ gerbv_gdk_draw_prim1(GdkPixmap *pixmap, GdkGC *gc, gerbv_simplified_amacro_t *s,
     const int diameter_idx = 1;
     const int x_offset_idx = 2;
     const int y_offset_idx = 3;
+    const int rotation_idx = 4;
     const gint full_circle = 23360;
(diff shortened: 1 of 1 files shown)


spe-ciellt:
I tested it and all renderers render it correctly. Good work!

PR #377, opened 2026-03-07
test: add libgerbv consumer link test and example program tests

rampageservices:
## Summary

- Adds a minimal C test program that links against `libgerbv` as a downstream consumer would, exercising `gerbv_create_project`/`gerbv_destroy_project` and forcing resolution of the dxflib `gerbv_export_dxf_file_from_image` symbol
- Compiles and runs the six documented example programs from `doc/example-code/` as integration tests, catching API drift in the examples that downstream developers rely on
- Examples 1-5 run headless (parse, export, draw); example 6 is compile-only since it opens a GTK window
- Changes examples 1-5 `main()` signature from `(int argc, char *argv[])` to `(void)` to fix `-Wunused-parameter` errors under `-Werror`

Since `cmake --build` builds all targets by default, these tests run in every CI configuration without any workflow changes.

## Test plan

- [ ] Verify `test_libgerbv` and `example1`-`example6` all build on CI targets
- [ ] Verify `libgerbv_link_test` and `libgerbv_example1`-`libgerbv_example5` pass in ctest
- [ ] Verify `example6` compiles but is not included as a ctest target
- [ ] Confirm no output files are left tracked in the source tree (`.gitignore` covers `*-output.*`)

Refs: #366


rampageservices:
### Why change `main()` to `main(void)` in examples 1-5?

The project's CMake base preset sets `CMAKE_COMPILE_WARNING_AS_ERROR=ON`, which promotes all warnings to errors. Examples 1-5 declare `int main(int argc, char *argv[])` but never reference `argc` or `argv`, so GCC's `-Wunused-parameter` fires and `-Werror` turns it into a build failure.

Two options to fix this:
1. Add `(void)argc; (void)argv;` casts at the top of each `main()`
2. Change the signature to `int main(void)`

I went with option 2 because these examples genuinely don't use command-line arguments — `int main(void)` is the correct C prototype when parameters aren't needed (C11 §5.1.2.2.1 defines both `int main(void)` and `int main(int argc, char *argv[])` as valid hosted entry points). It's a more accurate declaration of the interface than suppressing warnings about parameters that exist but are ignored.

Example 6 keeps `(int argc, char *argv[])` because it passes them to `gtk_init(&argc, &argv)`.

spe-ciellt:
## Overall analysis

Generally a good thing. Not sure how it affects the pipeline timing. If the time increases too much we have to start to parallelize more.

The `main()` thing above is a valid C signature and is permitted in the standard.

## Needs rebase 

There might be issues discovered during rebase, some of them listed below.

## Issues Found

### 1. Wrong library target name — `gerbv` does not exist after PR #425

```cmake
target_link_libraries(test_libgerbv PRIVATE gerbv PkgConfig::GLIB PkgConfig::GTK)
...
target_link_libraries(example${n} PRIVATE gerbv PkgConfig::GLIB PkgConfig::GTK)
```

PR #425 (merged into `develop`) renamed the library targets to `gerbv-shared`
and `gerbv-static`. There is no `gerbv` target or alias in
`src/CMakeLists.txt`. CMake will error at configure time:

```
CMake Error: add_executable called with target "test_libgerbv" which requires
target "gerbv" that is not in any directory in the build system.
```

Fix: replace `gerbv` with `gerbv-shared` in both `target_link_libraries` calls.
This is the root cause of the "needs rebase" status.

### 2. New ctest targets missing `LD_LIBRARY_PATH` / `PATH` environment

The existing `gerbv_regression_tests` correctly sets:

```cmake
set_tests_properties(gerbv_regression_tests PROPERTIES
    ENVIRONMENT_MODIFICATION "LD_LIBRARY_PATH=path_list_prepend:$<TARGET_FILE_DIR:gerbv-shared>"
)
```

None of the five new ctest targets (`libgerbv_link_test`,
`libgerbv_example1`–`libgerbv_example5`) set this. At runtime, the test
process cannot find `libgerbv.so.1` (it lives in `build/src/Debug/` or
`build/src/Release/`, not in a standard linker search path), and every test
will fail with:

```
error while loading shared libraries: libgerbv.so.1: cannot open shared object file
```

The same applies on Windows (MSYS2 CI) where `PATH` must be prepended with
the DLL directory. The existing test uses a `WIN32` guard for this:

```cmake
if(WIN32)
    set_tests_properties(... PROPERTIES ENVIRONMENT_MODIFICATION
        "PATH=path_list_prepend:$<TARGET_FILE_DIR:gerbv-shared>")
else()
    set_tests_properties(... PROPERTIES ENVIRONMENT_MODIFICATION
        "LD_LIBRARY_PATH=path_list_prepend:$<TARGET_FILE_DIR:gerbv-shared>")
endif()
```

The new tests need the same treatment, applied either individually or via a
helper that sets them for all six new targets at once.

### 3. Function pointer cast to `void *` is undefined behaviour in C

```c
volatile void *dxf_export = (void *)gerbv_export_dxf_file_from_image;
```

Casting a function pointer to an object pointer (`void *`) is undefined
behaviour in C89/C99/C11 (it is permitted by POSIX, and gcc/clang accept it
in practice, but `-Wpedantic` will warn). A conforming and equally effective
alternative that stays within C:

```c
/* Force linker to resolve the DXF export symbol (issue #366) */
void (*volatile dxf_check)(void) = (void (*)(void))gerbv_export_dxf_file_from_image;
(void)dxf_check;
```

Casting between function pointer types is defined by the C standard as long as
the function is not called through the wrong type. This is maintenance-level —
the test works on all targeted platforms as written — but worth fixing to keep
the codebase clean under strict compilation modes.

rampageservices:
@spe-ciellt Rebased and all three issues addressed in 1b3f343:

1. **Library target** — replaced `gerbv` with `gerbv-shared` in all `target_link_libraries` calls (fixes #425 rename).

2. **LD_LIBRARY_PATH/PATH** — extracted the platform-specific env mod into a `LIBGERBV_ENV_MOD` variable and applied it to all 7 test targets (regression + link test + examples 1-5). Uses the same `ENVIRONMENT_MODIFICATION` pattern as the existing regression test.

3. **Function pointer cast** — replaced `(void *)gerbv_export_dxf_file_from_image` with `(void (*)(void))gerbv_export_dxf_file_from_image`. Casting between function pointer types is well-defined in C; casting to `void*` is only guaranteed by POSIX.

All 6 new tests pass locally.

eyal0:
It's too bad that none of these will be noticed in CI.  :-(

rampageservices:
Good point. The CI workflow currently only runs the shell-based regression tests — it doesn't call `ctest` at all, so neither the new link test nor the example program tests would run in CI as things stand.

Adding a `ctest --test-dir build` step to the workflow would cover these. Happy to put up a PR for that if there's interest.

eyal0:
https://github.com/gerbv/gerbv/blob/develop/.github/workflows/ci.yaml#L97-L98

ctest is called but the results are ignored.

Here's my attempt to re-enable the unit tests: https://github.com/gerbv/gerbv/pull/427

PR #378, opened 2026-03-07
ci: add native Windows build with MSYS2 UCRT64

rampageservices:
## Summary

- Add a `ci-windows` GitHub Actions job that builds and tests gerbv natively on `windows-latest` using MSYS2 UCRT64 GCC
- New toolchain and preset files for the MSYS2 UCRT64 environment
- Guard `-mwindows` linker flag behind a GCC compiler check to prepare for future MSVC support
- Gate the release job on the new Windows CI alongside Ubuntu and macOS

## Motivation

The existing MinGW-w64 cross-compilation (via mini-cross on Fedora) proves gerbv compiles *for* Windows, but never runs the result. This job compiles and tests *on* Windows, catching platform-specific issues early.

## Test plan

- [x] Existing CI jobs (Ubuntu, macOS, Fedora, Debian, MinGW cross-compile) continue to pass
- [x] New `ci-windows` job configures, builds, and attempts tests on `windows-latest`
- [x] Regression test image diffs on Windows are expected (different rendering backend) — `continue-on-error: true` handles this


rampageservices:
@eyal0 Could you review this when you get a chance? It adds a native Windows CI job using MSYS2 UCRT64 to complement the existing MinGW cross-compilation.

rampageservices:
Worth noting: the `-mwindows` guard (`CMAKE_C_COMPILER_ID STREQUAL "GNU"`) is intentionally narrow. For a future MSVC toolchain, the equivalent would be `/SUBSYSTEM:WINDOWS` + `/ENTRY:mainCRTStartup` (or a `WinMain` wrapper), and the `compilation-flags` interface library would need generator expressions to map GCC flags to MSVC equivalents (`/W4`, `/O2`, `/wd4996`, `/Gy` + `/OPT:REF`, etc.). That's out of scope here since vcpkg doesn't have GTK2, but the guard avoids breaking anything if someone tries it before the GTK3 migration.

eyal0:
Windows build is failing on M_PI.  Explanation here: https://github.com/gerbv/gerbv/pull/368#issuecomment-4015114629

spe-ciellt:
Interesting platform. I am not sure what this does compared to the current windows build. Also, it currently needs a rebase and it seems it didn't pass the pipeline the first time (haven't looked into the details of it).

Also a more generic question. When there is a "test plan" attached to the description and not all the boxes have been ticked, is it considered "Work In Progress" or is it up to me do the tests?

- If the tests are done, please tick them in.
- If it is "Work In Progress", please set the status to "Draft".
- If it is something I should do, please do tell in the description.

The more complicated the PR is and/or if there is work I should do before merging, the longer it might take for me to merge it.

rampageservices:
@spe-ciellt Good point on the test plan — I've checked all three boxes now. The tests were passing on Ubuntu and macOS before, and the Windows job was the only failure (due to the `M_PI` issue eyal0 flagged). That's now fixed by #393 which just merged — I've merged develop into this branch to pick it up.

To your question about what this adds vs the existing Windows build: the current `ci-windows_amd64-cross` job cross-compiles using MinGW on Linux (via mini-cross). It produces a Windows binary but never executes it — so it only proves the code compiles for Windows, not that it runs. This new `ci-windows` job builds and runs the tests natively on a Windows runner, which is how eyal0's `M_PI` issue was caught.

CI should be green now across all platforms. No action needed from you beyond review.

PR #380, opened 2026-03-07
fix: start new cairo subpath before drawing aperture holes

rampageservices:
## Summary

Without `cairo_new_sub_path()`, Cairo connects the hole subpath to the outer aperture shape with an implicit line segment. This causes rendering artifacts on annular ring pads and any aperture with a drilled hole.

- Add `cairo_new_sub_path()` call in `gerbv_draw_aperture_hole()` before drawing the hole shape, so the hole is rendered as an independent subpath
- Regenerate 10 affected golden test images

## Attribution

Extracted from #163 ([@meantaipan](https://github.com/meantaipan)).

## Test plan

- [x] `cmake --preset linux-gnu-gcc && cmake --build build` — clean build
- [x] `ctest -C Debug` — 94/105 pass (11 pre-existing mismatches, unchanged from develop)
- [x] 10 golden images regenerated for tests that now render correctly


spe-ciellt:
## Cosmetic

`gerbv_draw_polygon` has a now-redundant workaround with a misleading comment

```c
/* skip first point, since we've moved there already */
/* include last point, since we may be drawing an aperture hole next
   and cairo may not correctly close the path itself */
for (i = 1; i <= (int)numberOfSidesInteger; i++) {
    ...
    cairo_line_to(...);  /* deliberately repeats the first vertex */
}
```

This comment (and the loop bound `i <= numberOfSidesInteger` instead of
`i < numberOfSidesInteger`) exists precisely because of the bug this PR fixes:
the author previously closed the polygon back to the start vertex so that the
implicit line Cairo would draw to the hole-arc start would be zero-length and
invisible. With `cairo_new_sub_path()` now in place, the workaround is no
longer needed and the comment is misleading.

The loop bound change would be a separate correctness question (closing vs.
not closing the polygon explicitly), but the comment should at minimum be
removed or updated so future readers are not confused about why the polygon is
drawn the way it is. Cosmetic, but likely to cause confusion in code review
later.

PR #232, opened 2024-01-08
pick & place: support for semi-automatic tools, protocol, part list view, translation of coordinates

kb-01:
Please see for reference
https://github.com/gerbv/gerbv/issues/230

I added also example project of a proof-of-concept board I designed long time ago which I used to assess e.g. accuracy of p&p nozzle tracking and translation to pcb coordinates on the screen.

The example can be opened like this:
/src/run_gerbv  -s /gerbv-pnp-sock -p example/pick-and-place-2/if2qs-top.proj 

Screenshot at 2024-01-08 18-54-24
Screenshot at 2024-01-08 18-54-38
Screenshot at 2024-01-08 18-55-15

PR #384, opened 2026-03-07
Fix memory leaks in layer unload, attribute destroy, and project handling

rampageservices:
Fixes four memory leaks found while reviewing #232:

- **`gerbv_unload_layer()`** — After `gerbv_destroy_fileinfo()`, the `fileinfo` struct pointer itself was never freed. `gerbv_destroy_project()` already does both `gerbv_destroy_fileinfo()` + `g_free()`, so this aligns the pattern.
- **`gerbv_attribute_destroy_HID_attribute()`** — The loop frees `default_val.str_value` for string attributes but never frees `name`, which is allocated via `strdup()` at `project.c:727`.
- **`read_project_file()`** — `initfile` is allocated by `gerb_find_file()` but never freed on either the error return path or the normal path after `g_fopen` succeeds.
- **`project_destroy_project_list()`** — The loop frees each node's `filename` and attributes but never frees the node struct itself.

These are standalone bugfixes extracted from @kb-01's work in #232 to make them easier to review and merge independently.

Co-authored-by: kb-01 <kb@sysmikro.com.pl>

### DIFF
--- a/src/gerbv.c
+++ b/src/gerbv.c
@@ -360,7 +360,8 @@ gerbv_unload_layer(gerbv_project_t *gerbvProject, int index)
     gint i;
 
     gerbv_destroy_fileinfo (gerbvProject->file[index]);
-    
+    g_free (gerbvProject->file[index]);
+
     /* slide all later layers down to fill the empty slot */
     for (i=index; i<(gerbvProject->last_loaded); i++) {
 	gerbvProject->file[i]=gerbvProject->file[i+1];
@@ -1077,12 +1078,15 @@ gerbv_attribute_destroy_HID_attribute (gerbv_HID_Attribute *attributeList, int n
 {
   int i;
 
-  /* free the string attributes */
+  /* free the string attributes and names */
   for (i = 0 ; i < n_attr ; i++) {
     if ( (attributeList[i].type == HID_String ||
 	  attributeList[i].type == HID_Label) &&
 	attributeList[i].default_val.str_value != NULL) {
       free (attributeList[i].default_val.str_value);
+    }
+    if (attributeList[i].name != NULL) {
+      free (attributeList[i].name);
     }
   }
 
@@ -1106,20 +1110,17 @@ gerbv_attribute_dup (gerbv_HID_Attribute *attributeList, int n_attr)
     exit (1);
   }
 
-  /* copy the attribute list being sure to strdup the strings */
+  /* deep-copy every entry: struct copy first, then strdup name and
+   * string values so the copy owns all its memory independently */
   for (i = 0 ; i < n_attr ; i++) {
+    nl[i] = attributeList[i];
+    nl[i].name = attributeList[i].name ? strdup (attributeList[i].name) : NULL;
 
     if (attributeList[i].type == HID_String ||
 	attributeList[i].type == HID_Label) {
-
-      if (attributeList[i].default_val.str_value != NULL) {
-	nl[i].default_val.str_value = strdup (attributeList[i].default_val.str_value);
-      } else {
-	nl[i].default_val.str_value = NULL;
-      }
-
-    } else {
-      nl[i] = attributeList[i];
+      nl[i].default_val.str_value =
+	  attributeList[i].default_val.str_value
+	      ? strdup (attributeList[i].default_val.str_value) : NULL;
     }
   }
 
--- a/src/project.c
+++ b/src/project.c
@@ -1011,8 +1011,10 @@ read_project_file(char const* filename)
     if ((fd = g_fopen(initfile, "r")) == NULL) {
 	scheme_deinit(sc);
 	GERB_MESSAGE(_("Couldn't open %s (%s)"), initfile, strerror(errno));
+	g_free(initfile);
 	return NULL;
     }
+    g_free(initfile);
 
     /* Force gerbv to input decimals as dots */
     setlocale(LC_NUMERIC, "C");
@@ -1059,10 +1061,11 @@ project_destroy_project_list (project_list_t *projectList){
 	
 	for (tempP = projectList; tempP != NULL; ){
 		tempP2 = tempP->next;
-		
+
 		g_free (tempP->filename);
 		gerbv_attribute_destroy_HID_attribute (tempP->attr_list, tempP->n_attr);
 		tempP->attr_list = NULL;
+		g_free (tempP);
 		tempP = tempP2;
 	}
 }


kb-01:
looks good.

spe-ciellt:
## Issues Found

### 1. `free(name)` in `gerbv_attribute_destroy_HID_attribute()` causes UB/crash on drill image teardown — **Blocking**

The fix assumes that every `attributeList[i].name` was heap-allocated via `strdup()`. That is true for attributes built in `read_project_file()` (line 727 of `project.c`). However, `gerbv_attribute_destroy_HID_attribute()` is also called from **`gerbv_destroy_image()`** (`src/gerb_image.c:152`) to free `image->info->attr_list`. That attribute list is populated by `gerbv_attribute_dup()` of `drill_attribute_list` (a static array with string-literal `name` fields such as `N_("autodetect")`, `N_("units")`, etc.) in `src/drill.c:577–578`.

`gerbv_attribute_dup()` does a **shallow struct copy** for non-`HID_String`/`HID_Label` attributes:

```c
} else {
    nl[i] = attributeList[i];   /* copies the name pointer verbatim */
}
```

All four entries in `drill_attribute_list` are `HID_Boolean`, `HID_Enum`, or `HID_Integer` — so every shallow copy carries a `name` pointer into `.rodata` (a string literal). After the PR, every call to `gerbv_destroy_image()` on a drill image will call `free()` on a string-literal address — **undefined behaviour**, almost always a crash.

The `src/drill.c` code even has a FIXME acknowledging this gap (line 566–567):
```c
/* FIXME there should probably just be a function to copy an
   attribute list including using strdup as needed */
```

**Correct fix requires one of:**
- Fix `gerbv_attribute_dup()` to `strdup()` the `name` field for every entry (the FIXME's intent), then the `free(name)` in `gerbv_attribute_destroy_HID_attribute()` becomes safe everywhere.
- Or: move the `free(name)` out of the shared helper and into `project_destroy_project_list()` directly (only that site is guaranteed to have `strdup`-allocated names).

### 2. Double-free of `name` when project attrs are dup'd into a drill image — **Blocking**

When gerbv opens a drill file with a pre-existing project (the `reload && attr_list != NULL` branch, `drill.c:570`), `gerbv_attribute_dup(attr_list, n_attr)` copies the project's `attr_list` into `image->info->attr_list`. For `HID_Boolean`/`HID_Enum`/`HID_Integer` entries this is again a shallow copy, so `image->info->attr_list[i].name` and `attr_list[i].name` point to the **same heap string** (originally `strdup`-allocated in `read_project_file()`).

After the PR:
1. `project_destroy_project_list()` calls `gerbv_attribute_destroy_HID_attribute()` → frees `attr_list[i].name`.
2. `gerbv_destroy_image()` calls `gerbv_attribute_destroy_HID_attribute()` → frees the same pointer again.

This is a **double-free**, producing heap corruption and undefined behaviour.

Both issues 1 and 2 stem from `gerbv_attribute_dup()` not duplicating `name`. The fix to `gerbv_attribute_dup()` (strdup the name in every branch) resolves both.

### 3. Redundant `(void *)` cast in `free()` call — Cosmetic

```c
free ((void *)attributeList[i].name);
```

`name` is declared as `char *` in `gerbv_HID_Attribute` (not `const char *`), so the cast to `void *` is unnecessary. `free(attributeList[i].name)` is sufficient and more readable.

kb-01:
interesting, so the fix must be more complex. neither mine solution nor
Jose's is correct.

the workarounds you proposed seem to be good.

On Mon, 2026-03-16 at 10:09 -0700, Stefan Petersen wrote:
> spe-ciellt left a comment (gerbv/gerbv#384)
> 
> Issues Found
> > 1. free(name) in gerbv_attribute_destroy_HID_attribute() causes
UB/crash on drill image teardown — Blocking
> > > > > > > > > The fix assumes that every attributeList[i].name was heap-allocated
via strdup(). That is true for attributes built in
read_project_file() (line 727 of project.c). However,
gerbv_attribute_destroy_HID_attribute() is also called from
gerbv_destroy_image() (src/gerb_image.c:152) to free image->info-
>attr_list. That attribute list is populated by gerbv_attribute_dup()
of drill_attribute_list (a static array with string-literal name
fields such as N_("autodetect"), N_("units"), etc.) in
src/drill.c:577–578.
> > gerbv_attribute_dup() does a shallow struct copy for non-
HID_String/HID_Label attributes:
> } else {
> >     nl[i] = attributeList[i];   /* copies the name pointer verbatim
*/
> }
> 
> > > > > All four entries in drill_attribute_list are HID_Boolean, HID_Enum,
or HID_Integer — so every shallow copy carries a name pointer into
.rodata (a string literal). After the PR, every call to
gerbv_destroy_image() on a drill image will call free() on a string-
literal address — undefined behaviour, almost always a crash.
> > The src/drill.c code even has a FIXME acknowledging this gap (line
566–567):
> /* FIXME there should probably just be a function to copy an
>    attribute list including using strdup as needed */
> 
> Correct fix requires one of:
> 
> > > Fix gerbv_attribute_dup() to strdup() the name field for every entry
(the FIXME's intent), then the free(name) in
gerbv_attribute_destroy_HID_attribute() becomes safe everywhere.
> > > Or: move the free(name) out of the shared helper and into
project_destroy_project_list() directly (only that site is guaranteed
to have strdup-allocated names).
> 
> > 2. Double-free of name when project attrs are dup'd into a drill
image — Blocking
> > > > > > > When gerbv opens a drill file with a pre-existing project (the reload
&& attr_list != NULL branch, drill.c:570),
gerbv_attribute_dup(attr_list, n_attr) copies the project's attr_list
into image->info->attr_list. For HID_Boolean/HID_Enum/HID_Integer
entries this is again a shallow copy, so image->info-
>attr_list[i].name and attr_list[i].name point to the same heap
string (originally strdup-allocated in read_project_file()).
> After the PR:
> 
> > project_destroy_project_list() calls
gerbv_attribute_destroy_HID_attribute() → frees attr_list[i].name.
> > gerbv_destroy_image() calls gerbv_attribute_destroy_HID_attribute() →
frees the same pointer again.
> 
> > This is a double-free, producing heap corruption and undefined
behaviour.
> > > Both issues 1 and 2 stem from gerbv_attribute_dup() not duplicating
name. The fix to gerbv_attribute_dup() (strdup the name in every
branch) resolves both.
> 3. Redundant (void *) cast in free() call — Cosmetic
> free ((void *)attributeList[i].name);
> 
> > > name is declared as char * in gerbv_HID_Attribute (not const char *),
so the cast to void * is unnecessary. free(attributeList[i].name) is
sufficient and more readable.
> 
> —
> Reply to this email directly, view it on GitHub, or unsubscribe.
> > You are receiving this because you were mentioned.Message ID: <gerbv/
***@***.***>
> 
> [
> {
> > ***@***.***": "http://schema.org",
> ***@***.***": "EmailMessage",
> "potentialAction": {
> ***@***.***": "ViewAction",
> > "target": "https://github.com/gerbv/gerbv/pull/384#issuecomment-40692
35504",
> > "url": "https://github.com/gerbv/gerbv/pull/384#issuecomment-40692355
04",
> "name": "View Pull Request"
> },
> "description": "View this Pull Request on GitHub",
> "publisher": {
> ***@***.***": "Organization",
> "name": "GitHub",
> > "url": "https://github.com"
> }
> }
> ]
-- 
Krzysztof Blaszkowski

rampageservices:
@spe-ciellt @kb-01 Great catch — both blocking issues are fixed in 977eafe.

**Root cause fix:** `gerbv_attribute_dup()` now does a struct copy first, then `strdup(name)` for every entry (not just string/label types). This resolves the FIXME at drill.c:566 and makes it safe to `free(name)` in the destroy function everywhere:

```c
for (i = 0 ; i < n_attr ; i++) {
    nl[i] = attributeList[i];
    nl[i].name = attributeList[i].name ? strdup(attributeList[i].name) : NULL;

    if (attributeList[i].type == HID_String ||
        attributeList[i].type == HID_Label) {
      nl[i].default_val.str_value =
          attributeList[i].default_val.str_value
              ? strdup(attributeList[i].default_val.str_value) : NULL;
    }
}
```

This fixes:
1. **Crash on drill image teardown** — `drill_attribute_list` names are string literals in `.rodata`; now they're strdup'd by `gerbv_attribute_dup()`, so `free()` is safe.
2. **Double-free** — project attrs dup'd into drill images no longer share name pointers; each copy owns its own.
3. **Cosmetic** — removed the `(void *)` cast on `free(name)`.

Also rebased onto current develop (clean merge, dropped the stale hunks from pre-#381/#410).

Issue #158, opened 2022-12-02
Enhancement request: more flexible Pick and Place input reader.


meantaipan:
The current PnP CSV reader seems to be specific to a particular column format.  Since CSV input is notoriously non-standard and hard to parse reliably, suggest deprecating the current PnP input reader and replace with ability to call an external program.  This means that the detailed implementation of a CSV reader is not part of gerbv, which will simplify maintenance.

To process PnP input, gerbv command line would have another option:

```
gerbv --pnp-reader="my-converter-script.py -extra -args"
```

```my-converter-script.py``` (or whatever) would be a translator program that converts the PnP CSV file into a standard canonical format that is easy for gerbv to parse.  If no ```--pnp-reader``` option is specified, then it uses the current reader for backwards compatibility.

Converter program would be called via system(3) or similar, with input and output file names as args, plus extra args:

```
my-converter-script.py  -extra -args  "<infile.csv>" tmp/gerbv-std-pnp.bin 
```

Something like that, anyway.

mikhailmihalkov:
Excuse me, why convert it? This file is usually loaded into the component installer program and is already being edited there....

meantaipan:
@mikhailmihalkov It's nice to be able to view component placement on top of the Gerber images.  This converter program is used so that random format PnP files can be viewed in gerbv.  It is already in there, but either make it work or take it out completely.

KjellMorgenstern:
I did some work on this here: https://github.com/KjellMorgenstern/gerbv/tree/refactor_pnp

Still work in progress, but it it works with the examples from gerbv (protel-pnp and pick-and-place subdir) and fritzings .xy export.

It tries to find a line that looks like a header, e.g.
`# RefDes, Description, Value, X, Y, rotation, top/bottom`
or
`RefDes,Description,Package,X,Y,Rotation,Side,Mount`

further tries to guess if SMD sizes are in Description, Package or Footprint column.

Issue #119, opened 2022-05-21
pick-and-place parser tries parsing comments


ooxi:
The current pick-and-place parsing implementation tries parsing comments and only fails through a coordinate sanity check instead of skipping comments (rows beginning with `%`) alltogether.

See [LED.xy](https://github.com/gerbv/gerbv/blob/main/example/pick-and-place/LED.xy) example.

rampageservices:
#385 improves detection robustness in `pick_and_place_check_file_type()` by adding position-aware matching for the Excellon markers (`G54`, `M00`, `M02`, `G02`, `ADD`). Previously these were matched anywhere in a line via `g_strstr_len()`, causing false positives when component values contained these substrings. The fix only flags these markers when they appear at the start of a line (position 0 or 1), where valid Excellon commands always appear.

This doesn't address the `%`-prefix comment parsing issue raised in the original report. @KjellMorgenstern's `refactor_pnp` branch takes a different approach with header detection and CSV parsing refactoring. Would it make sense to also add explicit `%`-prefix comment-line skipping as a complementary fix, or is the `refactor_pnp` approach preferred for that?

PR #385, opened 2026-03-07
Fix false-positive Excellon detection in PnP file type checker

rampageservices:
`pick_and_place_check_file_type()` searches for Excellon markers (`G54`, `M00`, `M02`, `G02`, `ADD`) anywhere in each line using `g_strstr_len()`. Component values containing these substrings (e.g., a part value of "M02" or "ADD1085") cause valid PnP files to be misidentified as drill/gerber files, since those markers make the function return `FALSE`.

This adds position checking so these markers are only recognized when they appear at the start of a line (position 0 or 1), where valid Excellon commands always appear. PnP data has these strings embedded in later CSV columns, so checking the position eliminates the false positives.

This partially addresses #119 by improving detection robustness, though it doesn't add explicit `%`-prefix comment skipping (as @KjellMorgenstern's `refactor_pnp` branch approaches). See also #158.

This is a standalone bugfix extracted from @kb-01's work in #232 to make it easier to review and merge independently.

Co-authored-by: kb-01 <kb@sysmikro.com.pl>

### DIFF
--- a/src/pick-and-place.c
+++ b/src/pick-and-place.c
@@ -432,6 +432,7 @@ pick_and_place_check_file_type(gerb_file_t *fd, gboolean *returnFoundBinary)
     int len = 0;
     int i;
     char *letter;
+    char *tmp;
     gboolean found_binary = FALSE;
     gboolean found_G54 = FALSE;
     gboolean found_M0 = FALSE;
@@ -461,19 +462,19 @@ pick_and_place_check_file_type(gerb_file_t *fd, gboolean *returnFoundBinary)
 	    }
 	}
 	
-	if (g_strstr_len(buf, len, "G54")) {
+	if ((tmp = g_strstr_len(buf, len, "G54")) && (tmp - buf < 2)) {
 	    found_G54 = TRUE;
 	}
-	if (g_strstr_len(buf, len, "M00")) {
+	if ((tmp = g_strstr_len(buf, len, "M00")) && (tmp - buf < 2)) {
 	    found_M0 = TRUE;
 	}
-	if (g_strstr_len(buf, len, "M02")) {
+	if ((tmp = g_strstr_len(buf, len, "M02")) && (tmp - buf < 2)) {
 	    found_M2 = TRUE;
 	}
-	if (g_strstr_len(buf, len, "G02")) {
+	if ((tmp = g_strstr_len(buf, len, "G02")) && (tmp - buf < 2)) {
 	    found_G2 = TRUE;
 	}
-	if (g_strstr_len(buf, len, "ADD")) {
+	if ((tmp = g_strstr_len(buf, len, "ADD")) && (tmp - buf < 2)) {
 	    found_ADD = TRUE;
 	}
 	if (g_strstr_len(buf, len, ",")) {


kb-01:
My concern is "sequence point" caused by "(term1 && term2)" which use the same tmp pointer. We get used to be sure that evaluation order starts from term1 and such assumption should be true for gcc. 
I have had additional checks against value of "len".
Anyway if you are fine with this change then let it be merged

spe-ciellt:
### 1. Style inconsistency: omits explicit `!= NULL` in `if()` condition — Cosmetic

The existing code in the same function uses an explicit `!= NULL` when assigning inside `if()`:

```c
if ((letter = g_strstr_len(buf, len, "R")) != NULL) {
```

The PR uses an implicit boolean for the same pattern:

```c
if ((tmp = g_strstr_len(buf, len, "G54")) && (tmp - buf < 2)) {
```

Both are correct C, but the explicit `!= NULL` form is the established style in this function. For consistency, it would read:

```c
if ((tmp = g_strstr_len(buf, len, "G54")) != NULL && (tmp - buf < 2)) {
```

This is cosmetic — the implicit boolean test is fine for a `gchar *` return — but aligning with the surrounding style would reduce noise in future diffs.

### 2. No regression test for the false-positive case — Maintenance

The bug is specific and reproducible: a PnP file whose component values contain `M02`, `G02`, `G54`, `M00`, or `ADD` somewhere after the first column. The test suite already has `example_pick_and_place_LED` exercising the happy path; a complementary file with one of those strings in a value column would lock in this fix. Without it, a future change that inadvertently removes the position check would pass all tests silently.

A minimal test input could be as simple as the existing `LED.xy` with one component renamed to, say, `"ADD1085"` or `"M02-cap"`. Adding an entry to `test/tests.list` and a golden reference would complete the fix.

kb-01:
regarding #2 my genuine patch has a check on "len" value. It could be
better than Jose's fix.

regarding #1, well, that is someone's "taste" dependent..


On Mon, 2026-03-16 at 10:17 -0700, Stefan Petersen wrote:
> spe-ciellt left a comment (gerbv/gerbv#385)
> 
> > 1. Style inconsistency: omits explicit != NULL in if() condition —
Cosmetic
> > The existing code in the same function uses an explicit != NULL when
assigning inside if():
> if ((letter = g_strstr_len(buf, len, "R")) != NULL) {
> 
> The PR uses an implicit boolean for the same pattern:
> if ((tmp = g_strstr_len(buf, len, "G54")) && (tmp - buf < 2)) {
> 
> > Both are correct C, but the explicit != NULL form is the established
style in this function. For consistency, it would read:
> > if ((tmp = g_strstr_len(buf, len, "G54")) != NULL && (tmp - buf < 2))
{
> 
> > > This is cosmetic — the implicit boolean test is fine for a gchar *
return — but aligning with the surrounding style would reduce noise
in future diffs.
> 2. No regression test for the false-positive case — Maintenance
> > > > > > > The bug is specific and reproducible: a PnP file whose component
values contain M02, G02, G54, M00, or ADD somewhere after the first
column. The test suite already has example_pick_and_place_LED
exercising the happy path; a complementary file with one of those
strings in a value column would lock in this fix. Without it, a
future change that inadvertently removes the position check would
pass all tests silently.
> > > > A minimal test input could be as simple as the existing LED.xy with
one component renamed to, say, "ADD1085" or "M02-cap". Adding an
entry to test/tests.list and a golden reference would complete the
fix.
> 
> —
> Reply to this email directly, view it on GitHub, or unsubscribe.
> > You are receiving this because you were mentioned.Message ID: <gerbv/
***@***.***>
> 
> [
> {
> > ***@***.***": "http://schema.org",
> ***@***.***": "EmailMessage",
> "potentialAction": {
> ***@***.***": "ViewAction",
> > "target": "https://github.com/gerbv/gerbv/pull/385#issuecomment-40692
91272",
> > "url": "https://github.com/gerbv/gerbv/pull/385#issuecomment-40692912
72",
> "name": "View Pull Request"
> },
> "description": "View this Pull Request on GitHub",
> "publisher": {
> ***@***.***": "Organization",
> "name": "GitHub",
> > "url": "https://github.com"
> }
> }
> ]
-- 
Krzysztof Blaszkowski

kb-01:
I have had no time to look into the issue. Will double check this at
the end of week likely.
Years have lapsed since I designed these changes.

On Tue, 2026-03-24 at 11:29 -0700, Stefan Petersen wrote:
> 
> Merged #385 into develop.
> 
> 
> —
> Reply to this email directly, view it on GitHub, or unsubscribe.
> > You are receiving this because you were mentioned.Message ID: <gerbv/
***@***.***>
> [
> {
> > ***@***.***": "http://schema.org",
> ***@***.***": "EmailMessage",
> "potentialAction": {
> ***@***.***": "ViewAction",
> > > > > "target": "https://github.com/gerbv/gerbv/pull/385?email_source=notif
ications\u0026email_token=AGD7ISVT5YV5JFSXP6I2JTT4SLHZBA5CNFSNUABQM5U
WIORPF5TWS5BNNB2WEL2JONZXKZKFOZSW45CON52GSZTJMNQXI2LPNYXTEMZYGYZTANZX
GM3THJTSMVQXG33OU5WWK3TUNFXW5JLFOZSW45FLM5WWC2LML5RWY2LDNM#event-
23863077373",
> > > > > "url": "https://github.com/gerbv/gerbv/pull/385?email_source=notifica
tions\u0026email_token=AGD7ISVT5YV5JFSXP6I2JTT4SLHZBA5CNFSNUABQM5UWIO
RPF5TWS5BNNB2WEL2JONZXKZKFOZSW45CON52GSZTJMNQXI2LPNYXTEMZYGYZTANZXGM3
THJTSMVQXG33OU5WWK3TUNFXW5JLFOZSW45FLM5WWC2LML5RWY2LDNM#event-
23863077373",
> "name": "View Pull Request"
> },
> "description": "View this Pull Request on GitHub",
> "publisher": {
> ***@***.***": "Organization",
> "name": "GitHub",
> > "url": "https://github.com"
> }
> }
> ]
-- 
Krzysztof Blaszkowski

spe-ciellt:
Looked like a simple fix and hopefully it helps someone, so it is merged and closed.

PR #393, opened 2026-03-07
fix: Define M_PI, M_PI_2, and M_1_PI.

### LINKED ISSUE #392: M_PI is undefined.
This fix got lost in the shuffle.

`M_PI` is a non-standard constant that is sometimes defined in `math.h`.  Let's define it as needed because it isn't standard and standard compilers are failing to compile gerbv when this is in use.

https://github.com/gerbv/gerbv/pull/368/changes#diff-3802cbcdb0af8c0a04ec0fc67cab4d18dbed5f686c2b6cad1d836323c0b64bbdR81-R90

```
#ifndef M_PI
# define M_PI		3.14159265358979323846	/* pi */
#endif
#ifndef M_PI_2
# define M_PI_2		1.57079632679489661923	/* pi/2 */
#endif
#ifndef M_1_PI
# define M_1_PI		0.31830988618379067154	/* 1/pi */
#endif
```

eyal0:
They are non-standard and not necessarily defined already.

This fixes #392.

### DIFF
--- a/src/gerbv.h
+++ b/src/gerbv.h
@@ -78,6 +78,16 @@ For help with using the standalone Gerbv software, please refer to the man page
 # include <cairo.h>
 #endif
 
+#ifndef M_PI
+# define M_PI		3.14159265358979323846	/* pi */
+#endif
+#ifndef M_PI_2
+# define M_PI_2		1.57079632679489661923	/* pi/2 */
+#endif
+#ifndef M_1_PI
+# define M_1_PI		0.31830988618379067154	/* 1/pi */
+#endif
+
 #if defined(__cplusplus)
 extern "C" {
 #endif


rampageservices (approved):
LGTM. Values match glibc and MSVC's `_USE_MATH_DEFINES` definitions exactly:

| Constant | PR value | MSVC/glibc |
|----------|---------|------------|
| `M_PI` | 3.14159265358979323846 | 3.14159265358979323846 |
| `M_PI_2` | 1.57079632679489661923 | 1.57079632679489661923 |
| `M_1_PI` | 0.31830988618379067154 | 0.318309886183790671538 |

`M_1_PI` differs only in trailing precision (20 vs 21 digits) — both well beyond `double` range (~15–16 significant digits), so no practical difference.

`#ifndef` guards are the standard pattern. Placement in `gerbv.h` ensures coverage for the `DEG2RAD`/`RAD2DEG` macros defined just below, plus all transitive includers.

spe-ciellt:
Looks simple enough, I merge it.

PR #396, opened 2026-03-08
feat: implement LR (Load Rotation) command

rampageservices:
## Summary

Implements the `LR` (Load Rotation) command from the Gerber spec (section 4.9.4).

- Parses `%LR<angle>*%` — angle in degrees CCW, stored as radians
- Adds `gdouble rotation` field to `gerbv_netstate_t` (initialized to 0.0)
- Applies `cairo_rotate()` in `draw_apply_netstate_transformation()` after mirroring and before axis select

### Files changed

- `src/gerbv.h` — add `rotation` field to `gerbv_netstate_t`
- `src/gerb_image.c` — initialize `rotation = 0.0` in new netstates
- `src/gerber.c` — parse LR command
- `src/draw.c` — apply rotation in netstate transformation

## Test plan

- [x] `test-load-rotation.gbx` loads without errors
- [x] Visual check: L-shapes appear at 0, 45, and 90 degree rotations
- [x] Golden reference image added and regression test passes
- [x] Existing tests unaffected — 95 passed, 11 failed (pre-existing), 0 skipped


spe-ciellt:
# Review

Thank you for fixing this, this is some steps to get GerberX2 support I guess. The first two (#394 and #395) I merged after some simple merge conflicts. I then also fixed some simple merge conflicts inherited by merge #394 and #395. 

There are some issues though, where i find 1 and 2 most severe.

## Issues Found

### 1. `rotation = 0.0` in `gerbv_image_return_new_netstate()` discards active LR state

**Severity: Blocking.**

`gerbv_image_return_new_netstate()` currently does:

```c
*newState = *previousState;   /* full struct copy — correctly propagates rotation */
previousState->next = newState;
newState->scaleA = 1.0;
newState->scaleB = 1.0;
newState->rotation = 0.0;     /* PR adds this — immediately overwrites the copy */
newState->next = NULL;
```

The `*newState = *previousState` copy already propagates `rotation` correctly. The added
`rotation = 0.0` line immediately discards it. Every RS274X command that creates a new
netstate — `%LM*%`, `%LS*%`, `%MO IN*%`, `%MO MM*%`, `%AS*%` — will silently reset an
active `%LR*%` back to 0°.

Per the Gerber X2 spec, LR state persists until the next explicit `%LR*%` command. It must
be inherited across intermediate netstates exactly as `mirrorState`, `unit`, and `axisSelect`
are — none of those have reset lines in the constructor.

**Fix:** remove the `newState->rotation = 0.0` line from `gerbv_image_return_new_netstate()`.
Only the `case A2I('L','R')` handler should write to `rotation`.

Note: `scaleA = 1.0` / `scaleB = 1.0` have the same pre-existing bug — any new-netstate
command clears an active `%LS*%`. The PR should at minimum not repeat the error for a third
field.

### 2. Cairo transform order: rotation is applied before mirroring in coordinate space

**Severity: Blocking.**

The current `draw_apply_netstate_transformation()` call sequence after the PR:

```c
cairo_scale    (cairoTarget, state->scaleA, state->scaleB);   /* LS */
cairo_translate(cairoTarget, state->offsetA, state->offsetB); /* OF */
/* mirror switch: */
cairo_scale    (cairoTarget, -1, 1);   /* LM — e.g. FLIPA */
/* PR inserts here: */
cairo_rotate   (cairoTarget, state->rotation);                /* LR */
/* axis select */
```

Cairo uses **left-multiplication** (prepend): each call computes `CTM_new = new_M × CTM_old`.
A point P transforms to device as `P' = CTM × P`. Reading the CTM product right-to-left
gives the coordinate-space application order:

```
CTM = R_lr × M_lm × T_of × S_ls
P' = R_lr × M_lm × T_of × S_ls × P
```

Coordinate order applied to P: **Scale → Translate → Mirror → Rotate.**

The Gerber X2 spec (section 4.9) defines the transformation order as:
**Scale (LS) → Rotate (LR) → Mirror (LM).**

Rotation must be called **before** the mirror switch so that the CTM product is
`M_lm × R_lr × S_ls`:

```c
cairo_scale    (cairoTarget, state->scaleA, state->scaleB);
cairo_translate(cairoTarget, state->offsetA, state->offsetB);
cairo_rotate   (cairoTarget, state->rotation);   /* LR — correct position */
switch (state->mirrorState) { ... }              /* LM */
/* axis select */
```

For files using only `%LR*%` without `%LM*%`, the ordering is invisible — there is no
mirror to commute with. The test file `test-load-rotation.gbx` contains no `%LM*%`
commands, so the golden PNG cannot detect this. Any file combining `%LR*%` and `%LM*%`
will render incorrectly.

### 3. `DEG2RAD()` macro not used

**Severity: Cosmetic.**

All other degree-to-radian conversions in `gerber.c` use `DEG2RAD()` (defined at
`gerbv.h:127`), e.g. `state->layer->rotation = DEG2RAD(gerb_fgetdouble(fd))` at line 1879.
The PR should use:

```c
state->state->rotation = DEG2RAD(gerb_fgetdouble(fd));
```

rampageservices:
Thanks @spe-ciellt — all three addressed in `ccc91d2`.

### 1. `newState->rotation = 0.0` reset (blocking)

Dropped. The `*newState = *previousState` copy already propagates `rotation` correctly per the X2 spec ("persists until the next explicit `%LR*%`"), and the case `A2I('L','R')` handler is the only path that should write to it.

Left `scaleA = 1.0` / `scaleB = 1.0` as-is to keep this PR scoped to the LR fix you flagged — per your note "the PR should at minimum not repeat the error for a third field". Added an inline comment so the next reader knows the LS reset is a pre-existing bug awaiting its own change. Happy to fix it here too if you'd prefer one combined commit.

### 2. Cairo transform order (blocking)

Moved the `cairo_rotate(state->rotation)` block to precede the `switch (state->mirrorState)`. The Cairo call order is now `scale → translate → rotate → mirror`, which (under Cairo's left-multiplication) yields a CTM of `M_lm × R_lr × T_of × S_ls`. Reading the product right-to-left gives the spec's coordinate-space order: **Scale → Rotate → Mirror** (X2 §4.9).

Added a comment block explaining the call-order ↔ coordinate-order inversion so future readers don't have to re-derive it. The existing `test-load-rotation.gbx` doesn't combine `%LR*%` with `%LM*%` so its golden PNG is unchanged; a follow-up test would lock this down.

### 3. `DEG2RAD()` macro

Switched. Also fixed the trailing 2-space indent on the `break` to match the surrounding 8-space tab indent in `gerber.c`.

Build clean locally (`gerbv-static` target). The pre-existing `main.c:1302` `strncasecmp` and `lrealpath.c:98` `canonicalize_file_name` implicit-declaration warnings under strict `-Werror` are unrelated and present on `develop` too — separate cleanup.

spe-ciellt:
Thanks for fixes, merging now.

PR #397, opened 2026-03-08
feat: implement AB (Aperture Block) command

rampageservices:
## Summary

Implements the `AB` (Aperture Block) command from the Gerber spec (section 4.11).

An aperture block defines a reusable group of graphics commands that can be
flashed like any other aperture. The syntax is `%ABD<code>*%` ... graphics ... `%AB*%`.

### Parser

- `%ABD<code>*%` — opens a block definition: creates a `GERBV_APTYPE_BLOCK` aperture entry, allocates a separate net list, and redirects net creation into it
- `%AB*%` — closes the block: stores the recorded net list on the aperture and restores normal net chain
- Block apertures are registered in the D-code stats list (fixes "undefined D code" warnings)
- Block recording is managed via state fields in `gerb_state_t`

### Renderer

The block renderer is extracted into `draw_render_block_nets()`, a static helper that supports all content types and recursive nesting:

| Content type | How rendered |
|---|---|
| Linear draws (D01) | Round-cap line stroke |
| Arc strokes (G02/G03 CW/CCW) | `cairo_arc`/`cairo_arc_negative` with scale transform |
| Rectangle draw strokes (D01 + rect aperture) | Filled quad path |
| Polygon regions (G36/G37) | `draw_render_polygon_object()` |
| Shape flashes (circle, rect, oval, polygon) | Standard shape draw + fill |
| Macro flashes (AM) | `gerbv_draw_amacro()` |
| Nested block flashes (AB inside AB) | Recursive call to `draw_render_block_nets()` |

### Cleanup

- Block net lists are freed when the image is destroyed

### Files changed

- `src/gerbv.h` — add `GERBV_APTYPE_BLOCK` enum, `block_netlist` field
- `src/gerber.h` — add block recording state to `gerb_state_t`
- `src/gerber.c` — parse AB open/close, redirect nets during recording, register in D-code stats
- `src/draw.c` — `draw_render_block_nets()` helper for full content type rendering
- `src/gerbv.c` — add "block" to aperture type name table
- `src/gerb_image.c` — free block net lists on image destroy
- `test/inputs/test-aperture-block.gbx` — exercises all content types
- `test/golden/test-aperture-block.png` — golden reference image

## Test plan

- [x] `test-aperture-block.gbx` loads without errors or warnings
- [x] Visual check: all 6 block types render (linear, arc, polygon, macro, rect stroke, nested block)
- [x] Existing tests unaffected (`ctest` — 95 pass, 11 pre-existing failures, 0 skip)


spe-ciellt:
# Review

Thank you for this PR. I solved a simple merge conflict here as well.  There are some comments on the code that I would be very glad if you could have a look at.

The blocking one is most serious.

## Issues Found

### 1. Nested AB blocks silently corrupt both apertures (blocking)

The PR claims to support "recursive nesting" of block apertures, but the parser does not actually support it. If a file contains:

```
%ABD20*%
  ...
  %ABD21*%    <- inner block opened while in_block == TRUE
    ...
  %AB*%
  ...
%AB*%
```

When `%ABD21*%` is parsed, `parse_rs274x()` sets `state->in_block = TRUE` (already TRUE) and `state->block_aperture_num = 21`, overwriting the saved `20`. The post-call check in `gerber_parse_file_segment()` is:

```c
if ((state->in_block) && (state->saved_curr_net == NULL)) {
    state->saved_curr_net = curr_net;
    curr_net = state->block_netlist;
```

Because `saved_curr_net` is already non-NULL (it was set when D20 opened), the redirect does NOT fire. `curr_net` still points into D20's net list, and D21 is never allocated a `block_netlist` (the new `g_new0(gerbv_net_t, 1)` is stored in `state->block_netlist` but `curr_net` does not switch to it). All nets subsequently drawn go into D20 rather than D21. When `%AB*%` for D21 fires, D21 gets a zero-length block netlist head and `block_aperture_num` resets to 20's number — which then fires the outer close incorrectly.

The renderer's `GERBV_APTYPE_BLOCK` case in `draw_render_block_nets()` does call itself recursively, but the parser never actually produces a nested structure, so the recursive renderer path is never exercised.

**Options:** (a) Reject nested AB at parse time with an error message and `gerb_ungetc`/skip, which matches the common real-world Gerber generator practice of non-nested blocks; (b) implement a true stack for `(block_aperture_num, block_netlist, saved_curr_net)` tuples if nesting is desired. The current silent corruption is worse than either.

### 2. Spurious `draw_fill` after `GERBV_APTYPE_BLOCK` flash (maintenance)

In `draw_image_to_cairo_target()`, after `draw_render_block_nets()` returns, execution falls through to the unconditional `draw_fill()` at line 1455 of the original file (the shared fill after the aperture-type switch). In `draw_render_block_nets()` itself, after the recursive `GERBV_APTYPE_BLOCK` inner case, the function similarly calls `draw_fill()` at the end of the flash branch. By the time either `draw_fill()` is reached, all Cairo path segments have already been consumed by the internal strokes and fills inside `draw_render_block_nets()`. Calling `cairo_fill()` on an empty current path is harmless (Cairo is tolerant), but it is semantically incorrect and misleading. The fix is to add a `break` before the shared fill in `draw_image_to_cairo_target()` (same as is done for the `GERBV_APTYPE_MACRO` case) and to move the `draw_fill()` inside the flash branch of `draw_render_block_nets()` to only run for non-block shapes.

### 3. Flash rendering ignores aperture holes for circle, rectangle, oval, polygon (maintenance)

In `draw_render_block_nets()`, when flashing a `GERBV_APTYPE_CIRCLE`, the code calls:

```c
gerbv_draw_circle (cairoTarget,
    image->aperture[block_net->aperture]->parameter[0]);
```

The corresponding code in `draw_image_to_cairo_target()` also calls `gerbv_draw_aperture_hole(cairoTarget, p[1], p[2], pixelOutput)` immediately after, which renders the optional inner hole for the aperture. `draw_render_block_nets()` omits the aperture hole call for all four simple aperture types (CIRCLE, RECTANGLE, OVAL, POLYGON). Apertures with holes are legal in Gerber and are used in PCB board designs for padstacks. An aperture block flashing a holed circle will render a solid disc instead of an annulus.

### 4. Arc direction logic inverted relative to main renderer (maintenance)

In the arc stroke branch of `draw_render_block_nets()`:

```c
if (block_net->cirseg->angle2 > block_net->cirseg->angle1)
    cairo_arc (cairoTarget, 0.0, 0.0, 0.5, ...);
else
    cairo_arc_negative (cairoTarget, 0.0, 0.0, 0.5, ...);
```

In `draw_image_to_cairo_target()` the same condition uses the same logic (`angle2 > angle1` → `cairo_arc`, else `cairo_arc_negative`). This is consistent. However, `calc_cirseg_mq()` and `calc_cirseg_sq()` in `gerber.c` set `angle1` as the start angle and `angle2` as the end angle. For a CW arc (G02), `angle2 < angle1` when the arc sweeps clockwise, so `cairo_arc_negative` (which also sweeps clockwise in Cairo's coordinate system with Y-down) is chosen — correct. For a CCW arc (G03), `angle2 > angle1`, so `cairo_arc` is chosen — also correct. The logic matches. This item is informational only.

rampageservices:
Thanks @spe-ciellt — thorough review, all four addressed in `5dcc666`. Picking your option (a) for the nested-AB case since it's lower-risk and matches the real-world generator practice you noted.

### 1. Nested `%ABD*%` definitions (blocking)

Now rejected at parse time with a `GERBV_MESSAGE_ERROR`. The outer block keeps running and the inner definition is dropped (`ap_num` never gets allocated as an aperture). The PR description's "recursive nesting" claim was conflating two distinct cases — I should have been clearer:

- **Nested definitions** (`%ABD21*%` opened while inside `%ABD20*%`): now explicitly rejected. Not supported.
- **Nested flashes** (D25's body flashes D20 via `D03`): supported and exercised by the existing `test-aperture-block.gbx` (block D25 flashes D20). This is the recursive renderer path in `draw_render_block_nets()`.

If real-world files ever need nested definitions, growing the single `(block_aperture_num, block_netlist, saved_curr_net)` triple into a stack is a self-contained future change. The renderer is already ready for it (kept the `GERBV_APTYPE_BLOCK` case in `draw_render_block_nets()` with a comment that it's unreachable from real input today).

### 2. Spurious `draw_fill` after BLOCK / MACRO flash (maintenance)

Added a `skip_shared_fill` boolean local in `draw_image_to_cairo_target()`'s flash branch, set `TRUE` by both `GERBV_APTYPE_MACRO` and `GERBV_APTYPE_BLOCK` (both internally paint via `gerbv_draw_amacro` and `draw_render_block_nets` respectively). The trailing `draw_fill` now only runs for cases that built a path but didn't fill it (CIRCLE / RECT / OVAL / POLYGON). Same fix mirrored inside `draw_render_block_nets()`'s own flash branch.

I noticed MACRO had the same stale-call pattern that you flagged for BLOCK, so I fixed both rather than fixing only BLOCK and leaving the MACRO precedent.

### 3. Flash rendering ignored aperture holes (maintenance)

Added the missing `gerbv_draw_aperture_hole()` calls after `gerbv_draw_circle / rectangle / oblong / polygon` in `draw_render_block_nets()`, matching the parameter offsets the main `draw_image_to_cairo_target()` flash path uses (`p[1]/p[2]` for circle, `p[2]/p[3]` for rect/oval, `p[3]/p[4]` for polygon).

### 4. Arc direction logic — no change

Confirmed informational. The renderer's `angle2 > angle1` → `cairo_arc` / `else` → `cairo_arc_negative` matches `draw_image_to_cairo_target()`'s logic, and the `calc_cirseg_*` start/end semantics make CW/CCW come out right under Cairo's Y-down system.

The existing golden PNG should be unchanged by these fixes since none of the test blocks use holed apertures and the spurious `draw_fill` was a no-op on the empty path. If you'd like a new test exercising a holed circle / rect / oval / polygon flashed inside a block (to lock in fix #3), happy to add one in a follow-up — let me know whether you prefer it here or as a separate PR.

spe-ciellt:
Thanks for fixing this. Sorry about the rebase mishaps. Will merge this now.

PR #401, opened 2026-03-08
ci: fix Windows regression test runner

rampageservices:
## Summary

Follow-up to #378. Fixes the remaining cross-platform CI issues from the Windows regression test work.

### Changes

1. **Fix xgettext configure-time race** (`config/CMakeLists.txt`)
   - On clean CI checkouts, `Gettext_helpers.cmake` runs `xgettext` at configure time, but `authors.c`/`bugs.c` don't exist yet (they're generated at build time)
   - Add `execute_process()` to generate these files at configure time using the same `generate_headers.cmake` script — pure CMake `file()` commands, no external tools
   - The build-time `add_custom_command` stays for incremental rebuilds

2. **Fix IMv7 `convert` deprecation** (`test/run_tests.sh`)
   - All CI platforms ship ImageMagick v7 where `convert`, `compare`, etc. are deprecated in favor of `magick` subcommands
   - On Windows, bare `convert` also collides with the Windows `convert.exe` disk utility
   - Auto-detect IMv7 via `command -v magick` and prefix all IM tool invocations accordingly
   - `convert` is a special case — it was merged into `magick` itself, so it maps to bare `magick` (not `magick convert`, which still emits a deprecation warning)
   - Environment variable overrides (`IM_COMPARE`, etc.) still work

### Previous changes (from earlier commits)

- Add `imagemagick:p` to the MSYS2 UCRT64 package list in `ci.yaml`
- Use `PATH` instead of `LD_LIBRARY_PATH` on Windows in `test/CMakeLists.txt`

### Test results

- **94/105 tests pass on Windows** — 11 rendering differences from cairo-win32 vs cairo-xlib (expected, handled by `continue-on-error: true`)
- Ubuntu and macOS CI jobs remain unaffected

## Test plan

- [x] Windows CI job runs regression tests without `command not found` errors
- [x] 94/105 tests pass on Windows (11 rendering differences expected)
- [x] Ubuntu and macOS CI jobs remain unaffected
- [x] No xgettext warnings on clean configure
- [x] No IMv7 deprecation warnings on any platform

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -243,6 +243,7 @@ jobs:
           gtk2:p
           gtkmm:p
           gettext-tools:p
+          imagemagick:p
 
     - name: Configure
       run: cmake --preset msys2-ucrt64-gcc
--- a/config/CMakeLists.txt
+++ b/config/CMakeLists.txt
@@ -76,6 +76,17 @@ option(USE_INTERFACE "Use extension interface of Scheme interpreter" ON)
 configure_file(config.h.in config.h @ONLY)
 
 set(GENERATOR_SCRIPT_DIR ${CMAKE_SOURCE_DIR}/cmake/generate)
+
+# Generate at configure time so xgettext can find them
+execute_process(
+    COMMAND ${CMAKE_COMMAND}
+            -DINPUT_BUGS=${CMAKE_SOURCE_DIR}/BUGS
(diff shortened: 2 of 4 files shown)


spe-ciellt:
One maintenance note: the `execute_process` should add `COMMAND_ERROR_IS_FATAL ANY` so a configure-time failure is reported immediately rather than manifesting as a confusing xgettext error downstream.

I will merge this before release, with or without fix.

PR #404, opened 2026-03-08
fix: Add upload of artifacts for macos and windows-msys2.

spe-ciellt:
Goal is to have them in the release.


spe-ciellt:
This was a quick test that failed. I go back to think before I do anything.

PR #414, opened 2026-03-09
build: Define QUARTZ/WIN32 in CMakeLists.txt for Apple/Windows.

eyal0:
QUARTZ and WIN32 were previously set only in CMake presets, so CI (which uses presets) passed, but builds without presets did not get the correct macro definitions.

Define them in CMakeLists.txt so that *all* Apple and Windows builds get the right macros, whether or not presets are used.


eyal0:
@rampageservices Please take a look.

eyal0:
An example of the failure to compile without this fix is here: https://github.com/pcb2gcode/pcb2gcode/actions/runs/22842518297/job/66251462527

An example of success is here, in the "Install gerbv" step: https://github.com/pcb2gcode/pcb2gcode/actions/runs/22841858512/job/66249519399  You can see that it is pulling a fork of this repo.  That fork has the fix.

rampageservices (commented):
Thanks for the fix — the pcb2gcode build failure is a real issue and the approach here is correct. `if(APPLE)` and `if(WIN32)` are built-in CMake variables that work without toolchain files, so this properly handles the non-preset build path.

Two suggestions:

### 1. Linux `_GNU_SOURCE` is inconsistent after this PR

After this change, the define placement is:
- **macOS**: `_GNU_SOURCE QUARTZ` in CMakeLists.txt (removed from toolchain) ✓
- **Windows**: `_GNU_SOURCE WIN32` in CMakeLists.txt (removed from toolchain) ✓
- **Linux**: `_GNU_SOURCE` only in toolchain file (not in CMakeLists.txt) ✗

A Linux build without a preset still won't get `_GNU_SOURCE`. For consistency, consider adding:

```cmake
    target_compile_definitions(compilation-flags INTERFACE _GNU_SOURCE)
endif()
```

Or since `_GNU_SOURCE` is already set for both macOS and Windows, just set it unconditionally for all platforms — it's benign everywhere and simplifies the logic:

```cmake
target_compile_definitions(compilation-flags INTERFACE _GNU_SOURCE)
if(APPLE)
    target_compile_definitions(compilation-flags INTERFACE QUARTZ)
endif()
if(WIN32)
    target_compile_definitions(compilation-flags INTERFACE WIN32)
endif()
```

### 2. Consider keeping toolchain defines rather than removing them

The toolchain files removing `add_compile_definitions(...)` means they're no longer self-documenting about what platform defines they expect. Duplicate `-D` flags are a no-op in practice, so keeping the defines in both places is harmless and makes the toolchain files clearer about their intent. CMakeLists.txt becomes the authoritative source for non-preset builds, while toolchain files remain readable on their own.

Not a blocker either way — just a style preference.

eyal0:
Let's add `_GNU_SOURCE` all the time.  It is necessary for `M_PI` and all platforms are using that.

One issue with duplicating it into both places is if later on you want to remove it, when you delete it, it doesn't work because you did it twice.  I prefer to leave it in just CMakeLists.txt, where it must be, and let the presets inherit it.  If anything, we'd like the presets to be as small as possible.  Ideally, they don't exist and each platform is able to run cmake without any special cases.

This issue continues to hold up compilation of pcb2gcode in CI.  I think that anyone that downloads and tries to compile gerbv on windows and macos is going to have a headache.

eyal0:
Can we make progress on this?  I'm currently using my own private copy of gerbv right now because this code will not cmake correctly.

I think that it is also still broken for `clang` and I will fix that after this.

spe-ciellt:
Sorry, I will not merge this as it is right now. The `_GNU_SOURCE` part can be merged, but platform/toolchain specific in toplevel `CMakeLists.txt` is a no at the moment.

There was a reason I removed those lines and that has not changed as far as I can see. I have not studied your pipeline in the first place, but...
1. I do not understand why you need to recompile this in the first place
2. Why can't you use the presets from the start? Works well for everyone else.
3. Presets and toolchains are here to stay, and I will not do "hacks" to get around them.

If there is a preset that you need, and it is of general interest, then please submit that. Also if there can be a special build we can provide it is also OK.

So if you update this PR with only the `_GNU_SOURCE` part and do a `git push --force-with-lease`  I will merge this.

eyal0:
The reason that I need to compile is because I'm using the latest version of gerbv on the develop branch.  The one from `apt-get` is very old and doesn't work.

Most users will compile like this:

```
mkdir build
cd build
cmake ..
make -j 20
```

That will not work for users because they will have to know about the presets.  It's not standard.

With this PR, the user can choose to use presets or not use them and still everything will work well.

eyal0:
Not having this is makes the code less user-friendly.

eyal0:
Also, can you add a preset `linux-clang` and `windows-clang`?  By not making this fix, we have to chase all the various presets.

eyal0:
Here's what needs to be done to get gerbv to compile.

https://github.com/pcb2gcode/pcb2gcode/pull/797/changes

You can see that I need to branch on the operating system because the compilation is unfortunately not consistent between platforms without this fix.

Clang compiles with the gcc preset.

spe-ciellt:
I can look into the Clang vs GCC presets. What is wrong with the GCC presets in Clang?

I find presets are extremely more user friendly, especially if you have more than one preset.

I am currently writing a BUILD.md so there is something to use as reference.

spe-ciellt:
First, I recommend you install Ninja in your pipeline. 
Secondly, I don't know what your LOCAL_INSTALL_PATH is set to, but CMake should by default install into `/usr/local` if nothing is set.  I am not sure where it ends up in the Windows environment though.

Then change your patch
```
if [[ "${{ matrix.os }}" == "windows" ]]; then
    cmake --preset msys2-ucrt64-gcc -G "Unix Makefiles" -DCMAKE_BUILD_TYPE=Release .. -DCMAKE_INSTALL_PREFIX=${LOCAL_INSTALL_PATH}
else
    cmake --preset linux-gnu-gcc -G "Unix Makefiles" -DCMAKE_BUILD_TYPE=Release .. -DCMAKE_INSTALL_PREFIX=${LOCAL_INSTALL_PATH}
fi
make -j 20
make install
```
to something like
```
if [[ "${{ matrix.os }}" == "windows" ]]; then
    cmake --preset msys2-ucrt64-gcc
    cmake --build --preset msys2-ucrt64-gcc-release
else
    cmake --preset linux-gnu-gcc
    cmake --build --preset linux-gnu-gcc-release
fi
cmake --install build
```

And I am still not sure why you have to recompile it in your pipeline.

eyal0:
The local install path is to avoid needing sudo. I don't mind that part. My annoyance is at needing to have a switch on Windows versus Ubuntu. I use a single CI pipeline with some if-else as needed to switch on whether or not I have coverage, gcc vs clang, etc. I try to have as few if-else as possible, to keep things uniform. Gerbv is the only build that needs if-else on operating system because it's the only one that is unable to detect Windows vs Ubuntu and compile correctly.

I'm not advocating for removing the presets. They are still useful for CI. But most users will not find the presets and will have trouble compiling. Because most projects don't require presets to compile.

This PR allows the project to compile with or without presets. If the concern is that you don't want to add this into CmakeLists.txt then it could go into a different.cmake file.

Anyway, if I didn't install from sources, then how would I get gerbv? The version from apt-get is too old.

eyal0:
I'm just using the presets instead.  Not my favored solution.  :-(

PR #417, opened 2026-03-09
Build libgerbv in shared mode when configuring for install.

### LINKED ISSUE #416: V2.12.0 build doesn't create .so files. Older v2.11.1 build does create the .so files
When I build v2.12.0 there are no generated libgerbv.so files to link with other programs. There is only a libgerbv.a file.

For the meantime I have reverted back to v.2.11.1.

spe-ciellt:
The SHARED flag was removed, but it was said that it still built in shared by default. Now I have updated so that when running the configuration preset linux-gnu-gcc-install the shared libgerbv.so will be built. That is by setting the flag "BUILD_SHARED_LIBS" to "ON".

The configuration preset linux-gnu-gcc-install is used when compiling files to be used in packaging for release. So by default it will continue to build libgerbv.a all the other times.

See more in #416 

Closes #416

### DIFF
--- a/cmake/preset/linux-gnu-gcc.json
+++ b/cmake/preset/linux-gnu-gcc.json
@@ -11,7 +11,8 @@
             "name": "linux-gnu-gcc-install",
             "inherits": "linux-gnu-gcc",
             "cacheVariables": {
-                "CMAKE_INSTALL_PREFIX": "/opt/gerbv.github/"
+                "CMAKE_INSTALL_PREFIX": "/opt/gerbv.github/",
+                "BUILD_SHARED_LIBS": "ON"
             }
         }
     ],


rampageservices (commented):
The change is correct for what it does, but I think it's too narrow a fix for the problem described in #416.

### Root cause

Commit c9fd344 (PR #314) changed `add_library(gerbv SHARED)` → `add_library(gerbv)` to support static builds. The commit message said "Default behavior is unchanged (shared)", but that's not accurate — CMake's `BUILD_SHARED_LIBS` defaults to **OFF** when not set, so the default silently flipped from shared to static. That's why #416 reporters no longer get `.so` files.

### This PR only helps preset users

Setting `BUILD_SHARED_LIBS=ON` in the `linux-gnu-gcc-install` preset fixes it for `--preset linux-gnu-gcc-install`, but:

1. **Plain builds still break**: Anyone doing `cmake -B build && cmake --build build && cmake --install build` (without presets) still gets static-only. This is the most common way third-party projects and distro packagers build.

2. **Non-opt package presets are affected too**: The `deb` and `rpm` package presets use the plain `linux-gnu-gcc` configure preset, not `linux-gnu-gcc-install`, so they'd still package static libraries.

### Suggested alternative

Set the default in `CMakeLists.txt` itself, which is the standard CMake idiom for this:

```cmake
option(BUILD_SHARED_LIBS "Build shared libraries" ON)
```

This restores the v2.11.x default (shared) while still letting users pass `-DBUILD_SHARED_LIBS=OFF` for static builds — which was the original intent of PR #314. Preset files don't need to be changed at all.

If you specifically want the install preset to always force shared regardless of user flags, the preset change here is still useful on top of that. But the `option()` default in CMakeLists.txt is the more important fix since it catches all build paths.

spe-ciellt:
This was intended as a proof-of-concept. Hence not  merged. 

The biggest problem with this is that it only fixes one preset, not all. But presets are a thing that people must learn to live with, it is not going away and I will not allow "hacks" to solve specific problems in other builds. The only CMake "hacks" allowed will be to circumvent problems in CMake itself.

I read up a bit on CMake yesterday, so I have a plan that I must test. It will not duplicate any settings and will hopefully solve the problems. But it is so far only in theory.

rampageservices:
Appreciate the thoughtful approach here — glad to see a focus on doing things the right way rather than piling on hacks. That philosophy aligns with ours as well.

Looking forward to seeing what you come up with after testing. The `option(BUILD_SHARED_LIBS ...)` idea is just one path — if you have a cleaner plan that avoids duplicating settings across presets, even better.

spe-ciellt:
OK, this wasn't generic enough. Closing this and opening a new and improved PR.

PR #421, opened 2026-03-10
fix: GDK rounding for all shape primitives (#167)

### LINKED ISSUE #167: Gerbv incorrectly rounds aperture coordinates in fast render mode

rampageservices:
## Summary

Extends the fix from PR #173 (which corrected arc bounding-box rounding) to all remaining shape primitives in the GDK renderer:

- **`gerbv_gdk_draw_circle`** — `x - dia / 2` → `round(x - dia / 2.0)`
- **`gerbv_gdk_draw_prim1`** (macro circle) — same pattern
- **`gerbv_gdk_draw_prim6`** (moiré) — inline `gdk_draw_arc` call
- **`gerbv_gdk_draw_prim7`** (thermal) — inline `gdk_draw_arc` call
- **`gerbv_gdk_draw_rectangle`** — `>> 1` → `round(... / 2.0)`
- **`gerbv_gdk_draw_oval`** — `>> 1` → `round(... / 2.0)`

Integer division (`/ 2` or `>> 1`) truncates toward zero, causing the bounding-box top-left corner to be off by one pixel for odd-valued dimensions. This makes concentric apertures (e.g. a pad with a drilled hole) appear misaligned at low zoom levels when using the GDK fast renderer.

Using `round(... / 2.0)` (which maps to `ceil()` via the local macro) matches the fix already applied to `gerbv_gdk_draw_arc` in #173.

Fixes #167

## Test plan

- [x] `cmake --build` completes without warnings
- [x] `ctest` regression suite: same 11 pre-existing pixel-mismatch failures on both `develop` and this branch (no new failures)
- [ ] Visual: load a Gerber with THT pads at low zoom in GDK mode, confirm pad/hole concentricity


rampageservices:
Closing — opened against wrong base branch. Reopening against `develop`.

Issue #290, opened 2026-02-27
Proposal: Add MAE tolerance threshold to regression test suite


rampageservices:
## Problem

The test suite (`test/run_tests.sh`) requires exact pixel match (MAE == 0) when comparing rendered output to golden reference PNGs. This causes false failures on any system with a different Cairo, fontconfig, or FreeType version than the one used to generate the golden files.

As noted during review of PR #282, 10 of the 101 tests fail on `develop` due to pre-existing Cairo-version mismatches — not regressions. The same 10 failures appear consistently across PRs (#282, #283, #285, #286).

### Current workarounds (both inadequate)

1. **`GERBV_MAGIC_TEST_SKIP=yes`** — skips ALL tests. Too coarse; hides real regressions.
2. **CI `continue-on-error: true`** — masks all failures. Defeats the purpose of the test suite.

## Proposal

Add a `GERBV_TEST_MAE_THRESHOLD` environment variable that sets the maximum acceptable Mean Absolute Error for pixel comparisons.

- **Default: `0`** — preserves current exact-match behavior
- **Non-zero value** — passes tests where MAE is within threshold, reports the actual MAE in output

### Changes to `test/run_tests.sh`

1. Read threshold: `mae_threshold=${GERBV_TEST_MAE_THRESHOLD:-0}`
2. Print threshold in the config banner
3. Replace the comparison logic to extract the numeric MAE value and compare against threshold
4. Update PASS message to show MAE when non-zero: `PASS (MAE=42 <= 500)`
5. Update FAIL message to show MAE and threshold: `FAILED (MAE=1200, threshold=500)`

### Changes to `test/run_drill_tests.sh`

Same threshold variable and comparison logic (ref #288 for eventual merge of these scripts).

### Usage

```bash
# Exact match (current default behavior)
./run_tests.sh

# Allow sub-pixel antialiasing differences
GERBV_TEST_MAE_THRESHOLD=500 ./run_tests.sh

# CI with tolerance for cross-platform Cairo differences
GERBV_TEST_MAE_THRESHOLD=500 ./run_tests.sh
```

### Pre-existing failures on `develop`

These 10 tests fail with MAE > 0 on current `develop` due to Cairo/font rendering differences, not code bugs:

| Test | Likely cause |
|------|-------------|
| `test-image-justify-*` | Sub-pixel antialiasing |
| `test-image-offset-*` | Viewport position rounding |
| `test-image-polarity-*` | Compositing differences |
[...]

PR #427, opened 2026-03-11
ci: Re-enable unit testing in CI.

eyal0:
Unit test failures are ignored in CI instead of being flagged as errors. This can cause PRs which cause a regression in the code to be inadvertently accepted.

There are currently some tests which fail to match the golden examples. There is also one test that causes gerbv to crash. The former have their golden files regenerated. The latter is commented out until the gerbv crash is resolved.


eyal0:
I can't believe that we were going this long without any unit testing!  Oh no!

eyal0:
I think that this is pretty straightforward.  Any progress on this one?

eyal0:
@spe-ciellt unit testing is broken. This fix ought to be prioritized so that future PRs will be checked for correctness before they get merged.

eyal0:
For example, this testing exposed a crash in gerbv that is for sure a bug. We would have caught it long ago if CI included testing.

spe-ciellt:
A misleading headline and no description. I am no mind reader. Your shouting doesn't help either.

Please, use reasonable headlines, add a description, do not just comment out code with "fix later"

There is no comment on what tests are failing and what error message or anything. The tests are just commented out.  I can not and will not spend time on detective work. Please be more specific. You say it crashes, but that is nothing I can confirm. Either cut-n-paste error messages or point to a build where this happens.

The #428 Just enables tests again, yet again confusing head line and no description.

Is it that you just remade the golden images? They have been wrong for while now due to differences in Cairo version between the golden images generation and what is actually used.

eyal0:
I added a description.

spe-ciellt:
See #290 to see the problem and potential fix.

eyal0:
Does this also fix the test which is crashing?

eyal0:
Unit testing continues to fail.

rampageservices:
I've reviewed this PR and here's my analysis.

## What this PR does right

1. **Removes `continue-on-error: true` from Linux CI** — this is the critical fix. Currently, test failures on Linux are silently ignored, which means regressions slip through. Making tests fail CI is essential.

2. **Adds `--output-on-failure`** — good addition, makes debugging CI failures much easier.

3. **Regenerates 5 stale golden images** — brings them in sync with current Cairo rendering.

## Issues

### 1. Commenting out the crashing test (`test-drill-trailing-zero-suppression`)

The test is commented out in `tests.list` with no explanation of what the crash is, what platform it occurs on, or a reference to a tracking issue. @spe-ciellt flagged this and it still hasn't been addressed. The test passes without crashing on Fedora 43 (I just verified), so the crash may be platform-specific.

At minimum, the comment should say *why* — e.g. `# Disabled: crashes on Ubuntu 22.04 (see #NNN)` with a filed bug. Better yet, investigate the crash root cause.

### 2. macOS and MSYS2 still have `continue-on-error: true`

These are left with TODO comments, which means regressions on those platforms remain invisible. This is understandable as a first step, but should be tracked.

### 3. Overlaps with #398 (MAE tolerance threshold)

PR #398 takes a more comprehensive approach to the same underlying problem:
- Adds a configurable `GERBV_TEST_MAE_THRESHOLD` (default 500) to `run_tests.sh` so sub-pixel anti-aliasing differences don't cause false failures
- Regenerates all 96 golden images (not just 5)
- Removes `continue-on-error: true` from **all** CI platforms (including macOS and Windows)
- Adds explicit font selection for PnP text rendering to reduce cross-platform font drift
- Addresses the root cause (#290) rather than just regenerating goldens that will drift again

If #398 lands first, this PR becomes unnecessary. If this PR lands first, #398 would need a rebase but still provides the MAE threshold which is the long-term fix for golden image drift.

### Recommendation

The immediate value here is removing `continue-on-error: true` on Linux — that alone is worth merging. But the commented-out test needs an explanation and ideally a tracking issue, and the stale golden regeneration is a partial fix that #398 does more thoroughly.

@eyal0 Could you add a comment in `tests.list` explaining the crash (platform, error message) and file a bug for it? That would address the maintainer's main objection.

eyal0:
@rampageservices I added the comment that you requested.

This PR is an improvement on the existing branch and #398 has been waiting for a whole month.  How about doing this one in the meanwhile?

PR #428, opened 2026-03-11
test: Enable test-drill-trailing-zero-suppression.

eyal0:


eyal0:
Because our CI was not actually running tests, this error was never noticed.  It causes a crash in gerbv.

After fixing the root cause of the crash, rebase this on top of the fix and CI should pass once again.

This changes depends on #427.  Commit #427 first.

PR #430, opened 2026-03-13
Moved install() command for init.scm to tinyscheme CMakeLists.txt

spe-ciellt:
A tiny fix to make sure all tinyscheme things (both files and configurations) is in the thirdparty/tinyscheme.

Took the opportunity to make sure some CMakeLists.txt is more readable by adding newlines.


spe-ciellt:
Must test if the binaries actually finds `init.scm` properly, not just that it is in the package.

PR #438, opened 2026-03-17
cmake: add COMMAND_ERROR_IS_FATAL ANY to configure-time execute_process calls

### LINKED ISSUE #403: The `execute_process` should add `COMMAND_ERROR_IS_FATAL ANY`
One maintenance note: the `execute_process` should add `COMMAND_ERROR_IS_FATAL ANY` so a configure-time failure is reported immediately rather than manifesting as a confusing xgettext error downstream.

_Originally posted by @spe-ciellt in https://github.com/gerbv/gerbv/issues/401#issuecomment-4019395278_

rampageservices:
Closes #403.

## Problem

The three `execute_process` calls that run at configure time have no failure handling. If any of them exit non-zero, CMake silently continues and the failure surfaces later as a confusing downstream error — the specific example in the issue is an `xgettext` error caused by the header-generation step failing to produce `bugs.c`/`authors.c`.

## Change

Add `COMMAND_ERROR_IS_FATAL ANY` to:
- `config/CMakeLists.txt` — the `generate_headers.cmake` script invocation
- `cmake/Gettext_helpers.cmake` — the `xgettext` `.pot` generation
- `cmake/Gettext_helpers.cmake` — the `msginit` `.po` initialisation

`cmake/GetGitVersion.cmake` is left unchanged; those calls already use `RESULT_VARIABLE` and are intentionally tolerant of `git` not being installed.

`COMMAND_ERROR_IS_FATAL` requires CMake ≥ 3.19; the project's `cmake_minimum_required` is 3.28, so there is no compatibility concern.

### DIFF
--- a/cmake/Gettext_helpers.cmake
+++ b/cmake/Gettext_helpers.cmake
@@ -101,7 +101,8 @@ function(configure_gettext)
             COMMAND "${GETTEXT_XGETTEXT_COMMAND}" ${GETTEXT_XGETTEXT_ARGS}
                 ${GETTEXT_SOURCES}
                 "--output=${GETTEXT_POTFILE_DESTINATION}/${GETTEXT_DOMAIN}.pot"
-            WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}")
+            WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
+            COMMAND_ERROR_IS_FATAL ANY)
     endif()
     add_custom_command(
         OUTPUT "${GETTEXT_POTFILE_DESTINATION}/${GETTEXT_DOMAIN}.pot"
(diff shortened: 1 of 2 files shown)


spe-ciellt:
Thanks for adding the flags. It exposed a very obvious problem, which I haven't noticed. I do not understand why this is happening in the pipeline, since it not happening locally. Needs examination.

I also noticed that when the build in the pipeline uses shallow copies of the repository and the version strings got all messy. So now I added #443 , so hopefully that solves the problem.

eyal0:
> Thanks for adding the flags. It exposed a very obvious problem, which I haven't noticed. I do not understand why this is happening in the pipeline, since it not happening locally. Needs examination.
> 
> I also noticed that when the build in the pipeline uses shallow copies of the repository and the version strings got all messy. So now I added #443 , so hopefully that solves the problem.

This is another dependency issue, yes? authors.c is a generated file and the build has a parallelism level of sixteen. If you want to reproduce it on your computer, you must remember to use an argument like `-j 16`.

I sent a PR with a similar fix perhaps for a different target.

spe-ciellt:
Which PR?

eyal0:
I think #365 

It forces authors.c to be made before the .pot files. I think that is the error in the ci build.

eyal0:
https://github.com/gerbv/gerbv/actions/runs/23174090248/job/67332350422?pr=438

Is that job failing to make the .pot file because authors.c is missing? I can see that in the environment variable, the parallelism is set to 16.

spe-ciellt:
The fix is easy hopefully. Swap the lines according to #444 . @rampageservices if you add the fix from #444 to this PR, and the compilation passes,  I will happily merge it . I will kill of the #444, just keeping it for reference until the fix is added.

spe-ciellt:
OK, I couldn't rebase as I thought. I will take a gamble to just merge this branch. I have tested it locally in all sorts of ways, so hopefully I will not break anything.

PR #444, opened 2026-03-17
Swap config stage and locale stage in CMakeLists.txt

spe-ciellt:
Experimental test


spe-ciellt:
Not valid, replaced by #445

PR #452, opened 2026-03-20
docs: GTK2 deprecation — plan migration from GdkColor to standalone color type

rampageservices:
## Summary

GTK2 is end-of-life. libgerbv's public API (`gerbv.h`) still depends on GTK2 via `#include <gtk/gtk.h>` and uses `GdkColor` (16-bit RGB, no alpha) in its public structs.

This PR adds documentation outlining the issue and a 3-phase migration plan.

## Problems

1. **`gerbv.h:73`** includes all of GTK2, even for headless/library-only usage
2. **`GdkColor`** in `gerbv_fileinfo_t.color` and `gerbv_project_t.background`:
   - No alpha channel (alpha is a separate `gushort` field)
   - 16-bit values (0-65535) instead of normalized floats
   - Deprecated in GTK3 in favor of `GdkRGBA`
3. **Cairo rendering** (`gerbv_render_layer_to_cairo_target`) doesn't need GTK at all, but the header forces linkage
4. **GLib 2.88+ warnings**: `GTypeDebugFlags` and `GTimeVal` deprecation warnings from GTK2 headers

### Phase 1: Standalone color type (non-breaking)
```c
typedef struct {
    double red;    /* 0.0 - 1.0 */
    double green;  /* 0.0 - 1.0 */
    double blue;   /* 0.0 - 1.0 */
    double alpha;  /* 0.0 - 1.0 */
} gerbv_color_t;
```
Replace `GdkColor` in public structs. Provide backward-compat macros.

### Phase 2: Remove GTK from gerbv.h
Move `#include <gtk/gtk.h>` into GUI sources only. Header needs only `<glib.h>` + `<cairo.h>`.

### Phase 3: Headless build option
`-DGERBV_HEADLESS=ON` to build libgerbv without GTK (common for server-side rendering).

## Context

We maintain a [fork](https://github.com/SourceParts/gerbv) and use libgerbv for server-side Gerber rendering (Cairo → PNG) in our electronics platform. The GTK2 dependency is the main friction point for headless/embedded usage.

### DIFF
--- a/GTK2_DEPRECATION.md
+++ b/GTK2_DEPRECATION.md
@@ -1,0 +1,82 @@
+# GTK2 Deprecation in libgerbv
+
+## Status
+
+GTK2 is **end-of-life** as of GTK3 (December 2020). GTK4 is the current stable release.
+libgerbv still depends on GTK2 for its GUI and uses `GdkColor` (16-bit RGB) in its public API.
+
+## Impact on libgerbv
+
+### API Surface Issues
+
+1. **`gerbv.h:73`** — `#include <gtk/gtk.h>` pulls in all of GTK2, even for headless/library-only usage
+2. **`GdkColor` in public structs** — `gerbv_fileinfo_t.color` and `gerbv_project_t.background` use `GdkColor` which:
+   - Has no alpha channel (alpha is a separate `gushort` field on `gerbv_fileinfo_t`)
+   - Uses 16-bit color values (0-65535) instead of normalized floats
+   - Is deprecated in GTK3 in favor of `GdkRGBA` (float RGBA)
+3. **`gerbv_render_layer_to_cairo_target()`** — the Cairo rendering path doesn't need GTK at all, but the header forces GTK2 linkage
+
+### Compilation Warnings (GLib 2.88+)
+
+```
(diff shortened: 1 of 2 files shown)


spe-ciellt:
This is a good plan, but I think really it should be assigned an Issue in the issue tracker. And then PR will reference that one, since I assume there will be more PRs like this. You can even create sub issues. It is great for everyone to keep track on how far this has come.

In the first place I am actually surprised (not now, but before) that libgerbv requires GTK. The thought was that libgerbv should only be parsing of Gerbers. Now I notice other things have crept in like exporters and other stuff, which then requires GTK.

So the ultimate would be to separate out all drawing (including export) from libgerbv to  a separate library.  But I imagine that would be a major undertaking, so that have to be improved on later.

I have not looked at the code yet, but I will.

rampageservices:
Good idea @spe-ciellt — I'll create a tracking issue for the GTK2 migration with sub-issues, and close this PR in favor of that. The PR can serve as the plan/reference, but the issue tracker is better for ongoing work.

Agreed on the libgerbv concern — parsing should ideally be separated from drawing/export. That's a bigger refactor but worth tracking.

PR #459, opened 2026-03-24
Fallback on CMake version if no git

### LINKED ISSUE #454: Version information not stored in the source code
Currently if I download the source code it is incomplete, it doesn't contain the version information. It defaults to v4.0, which is invalid. From the log during compile I see this
```
-- Could NOT find Git (missing: GIT_EXECUTABLE) 
-- git Version: v4.0
-- Version from git: v4.0
-- Parsed version: 4.0.0 (suffix: )
```
To compile one needs to obtain a complete copy of the gerbv repository in order to extract metadata from git to discover the version. Tens of megabytes of download for Just a few bytes of metadata.

This makes it redundant to download the source code, because another more complete copy (the repository history and metadata) needs to be downloaded anyway. This means the source code is not standalone, and is tethered to github, and git, to get the version.

spe-ciellt:
If Git is not installed the CMake process falls back to the version in `project()` in `CMakeLists.txt`.

Though we always build releases with using `git tag` we need to get the version into the GitHub source tar ball. How the source tarballs are created is not in our control. So we need to trick a version into there, but at the same time make sure all versions are the same. We now have two versions of truth and we need to make sure that they agree.

So now there is a little extra job that runs at release that checks that the versions are the same. The actual release job then depends on that the "check job" does not fail before starting the actual release process. That way we can hopefully avoid making a release where the two version numbers differs.

Closes #454

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -14,6 +14,24 @@ on:
 
 
 jobs:
+  check-version:
+    runs-on: ubuntu-22.04
+    if: startsWith(github.ref, 'refs/tags/v')
+    steps:
+      - uses: actions/checkout@v6
+      - name: Check CMakeLists.txt VERSION matches git tag
+        run: |
+          CMAKE_VERSION=$(grep -oP 'project\s*\(\s*gerbv\s+VERSION\s+\K[0-9]+\.[0-9]+\.[0-9]+' CMakeLists.txt)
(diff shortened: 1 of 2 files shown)


eyal0:
This PR has broken the windows build for pkg-config

spe-ciellt:
No, I guess it was actually #462 but there is PR #464 waiting for approval to fix this.

PR #464, opened 2026-03-31
Add back installation of libgerbv.pc in Windows since it is apparentl…

### LINKED ISSUE #463: This PR has broken the windows build for pkg-config
This PR has broken the windows build for pkg-config

_Originally posted by @eyal0 in https://github.com/gerbv/gerbv/issues/459#issuecomment-4164540645_

spe-ciellt:
…y used.

It was removed from Windows build in #462 since it is not common for Windows to use it.

Closes #463 .

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -106,7 +106,5 @@ set(includedir "\$\{prefix\}/${CMAKE_INSTALL_INCLUDEDIR}")
 set(PACKAGE ${PROJECT_NAME})
 configure_file(libgerbv.pc.in libgerbv.pc @ONLY)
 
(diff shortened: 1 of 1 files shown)


eyal0:
It's used in msys32, though.  Would you like to add a test to see that it's working?  In ci, you could run `pkg-config libgerbv`.  It should return true if libgerbv is correctly found.

eyal0:
To add the check for libgerbv.pc, you also need to install gerbv and not just build it in CI.

If you installed it, then you could run pkg-config libgerbv, which would exit will failure if libgerbv is not found.

Otherwise, this fix looks fine to me.  I didn't see #462 when the file was being removed.

PR #465, opened 2026-04-10
website: show actual package formats in download buttons

### LINKED ISSUE #451: Update website, especially with some new icons if possible
The website is now updated from the pipeline with the latest `develop` build. We build `.deb`, `.rpm`, etc, but the icons still show `.tar.gz` and `.zip`.

So we obvious would need some icons showing just `.deb`, `.rpm`, etc. Or we maybe could replace the icons with plain text?

Also is only one of the Windows builds listed and no MacOS build is listed.

rampageservices:
The download table on the website currently uses `.tar.gz` and `.zip` sprite icons (inherited from the Slate theme) for all download links, but the CI actually produces platform-specific packages:

| Platform | Actual format | Was showing |
|---|---|---|
| Fedora 43 | `.rpm` | `.tar.gz` sprite |
| Ubuntu 22.04 | `.deb` | `.tar.gz` sprite |
| Debian 13 | `.deb` | `.tar.gz` sprite |
| Windows amd64 | `.exe` (NSIS) | `.zip` sprite |
| Windows MSYS2 | `.zip` | `.zip` sprite (correct) |
| macOS | `.zip` | `.zip` sprite (correct, but no icon) |

This PR:

- Replaces the sprite-based `tar_download_link`/`zip_download_link` classes in the download table with a new `download_button` CSS class that shows the actual file extension as readable text (`.rpm`, `.deb`, `.exe`, `.zip`)
- Adds an Apple logo SVG for the macOS row (previously showed plain text "macOS" with no icon)
- Keeps the header source download sprites unchanged (those link to actual `.zip`/`.tar.gz` source archives)

Closes #451


spe-ciellt:
Good idea, better than what is there today. So I merge this without an deeper analysis.

PR #473, opened 2026-04-11
fix: enable destroy_paths() cleanup via atexit

### LINKED ISSUE #472: Memory leaks in init_paths() — static path strings never freed
## Description

`init_paths()` in `src/project.c` allocates four static strings (`bindir`, `exec_prefix`, `pkgdatadir`, `scmdatadir`) that are never freed, causing valgrind to report 181 bytes definitely lost.

Additionally, line 458 does `bindir = strdup(lrealpath(argv0))` — `lrealpath()` already returns a `malloc`'d string, so the `strdup` wrapper leaks the `lrealpath` result.

## Existing code

A `destroy_paths()` function already exists in the source but is `#if 0`'d out (lines 412-437) with a comment:

> *"this really should not be needed but it could be hooked in to appease malloc debuggers as we don't otherwise free these variables. However, they only get malloc-ed once ever so this is a fixed leak of a small size."*

## Valgrind output

```
==287282== 89 bytes in 1 blocks are definitely lost in loss record 224 of 254
==287282==    at 0x4840B26: malloc (vg_replace_malloc.c:447)
==287282==    by 0x418911: init_paths (project.c:527)
==287282==    by 0x418911: read_project_file (project.c:965)
==287282==    by 0x416E23: main_open_project_from_filename (main.c:231)

==287282== 92 bytes in 1 blocks are definitely lost in loss record 228 of 254
==287282==    at 0x4840B26: malloc (vg_replace_malloc.c:447)
==287282==    by 0x57D9A2E: strdup (strdup.c:42)
==287282==    by 0x576BAA9: realpath_stk (canonicalize.c:411)
==287282==    by 0x41B4F8: lrealpath (lrealpath.c:98)
==287282==    by 0x4188A8: init_paths (project.c:458)
```

## Fix

- Remove the `#if 0` guard on `destroy_paths()` and register it via `atexit()`
- Replace `strdup(lrealpath(argv0))` with `lrealpath(argv0)` to avoid the double allocation

PR: #473

rampageservices:
## Summary

Fixes #472 — static path strings in `init_paths()` (`src/project.c`) were never freed, causing valgrind to report 181 bytes definitely lost.

### Changes

1. **Enable `destroy_paths()`**: A cleanup function already existed but was `#if 0`'d out with a comment about "appeasing malloc debuggers." Removed the guard and registered it via `atexit()` so the paths are freed on normal exit.

2. **Fix double allocation**: `bindir = strdup(lrealpath(argv0))` leaked the `lrealpath()` return value since `lrealpath()` already returns a `malloc`'d string. Changed to `bindir = lrealpath(argv0)`.

### Verification

Valgrind before fix (via Xvfb with TinyScheme project loader):
```
definitely lost: 181 bytes in 2 blocks
```

Valgrind after fix:
```
definitely lost: 0 bytes in 0 blocks
```

Together with #470 (`gerbv_image_duplicate_image` leaks), gerbv now has **zero valgrind-reported leaks** across all tested code paths: Gerber parse/destroy, drill parse/destroy, PnP, aperture macros, multi-layer projects, RS274-X/DXF/drill export, and TinyScheme project file loading.

## Test plan

- [x] Builds clean with `-Werror`
- [x] No new test regressions (same 11 pre-existing golden-file mismatches)
- [x] Valgrind: 0 definitely lost bytes on TinyScheme project load path (Xvfb)
- [x] Valgrind: 0 definitely lost bytes on parse/destroy sweep

### DIFF
--- a/src/project.c
+++ b/src/project.c
@@ -392,15 +392,10 @@ static char *exec_prefix = NULL;
 static char *pkgdatadir = NULL;
 static gchar *scmdatadir = NULL;
 
-/* this really should not be needed but it could
- * be hooked in to appease malloc debuggers as
- * we don't otherwise free these variables.  However,
- * they only get malloc-ed once ever so this
- * is a fixed leak of a small size.
- */
-#if 0
-void
-destroy_paths ()
+/** Free static path strings allocated by init_paths().
+ * Registered via atexit() so valgrind reports zero leaks. */
+static void
+destroy_paths (void)
 {
   if (bindir != NULL) {
     free (bindir);
@@ -421,10 +416,7 @@ destroy_paths ()
     g_free (scmdatadir);
     scmdatadir = NULL;
   }
-
-
 }
-#endif
 
 static void
 init_paths (char *argv0)
@@ -455,7 +447,9 @@ init_paths (char *argv0)
   DPRINTF("%s (%s): haspath = %d\n", __FUNCTION__, argv0, haspath);
   if (haspath)
     {
-      bindir = strdup (lrealpath (argv0));
+      /* lrealpath() returns a malloc'd string — assign directly,
+       * don't strdup again (would leak the lrealpath result). */
+      bindir = lrealpath (argv0);
       found_bindir = 1;
     }
   else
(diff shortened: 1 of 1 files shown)


spe-ciellt:
Valgrind is a great tool! Merging this now.

Issue #483, opened 2026-05-14
Hello, how can I call this in VS2022 to parse Gerber files and generate images?


icewenwu:
Hello, how can I call this in VS2022 to parse Gerber files and generate images? window x64 ，c++ dll,thanks!

rampageservices:
Hi @icewenwu — short answer: gerbv ships a reusable C library (`libgerbv`), and you can call it from a VS 2022 C++ project, but gerbv itself only builds with GCC (MinGW-w64 / MSYS2 UCRT64). The supported path on MSVC is: build the DLL with MSYS2, then link to it from VS 2022 via a generated MSVC import library.

A minimal "parse Gerber → write PNG" looks like:

```cpp
extern "C" {
#include <gerbv.h>
}

int main() {
    gerbv_project_t *p = gerbv_create_project();
    gerbv_open_layer_from_filename(p, "input.gbr");
    gerbv_export_png_file_from_project_autoscaled(p, 800, 600, "out.png");
    gerbv_destroy_project(p);
}
```

The full set of steps — `cmake --preset msys2-ucrt64-gcc`, `gendef libgerbv-1.dll`, `lib /def:libgerbv-1.def /machine:x64 /out:libgerbv.lib`, include paths for the transitive GLib/GTK 2/Cairo headers, runtime DLLs to ship — is now written up in PR #484, which adds `doc/example-code/README.md`.

A couple of caveats worth knowing up front:
- `gerbv.h` transitively pulls in GLib and GTK 2 headers. Even for a "headless" PNG-render use case you still need those headers at compile time and the corresponding runtime DLLs at load time. A GLib/GTK-free build is not currently supported.
- libgerbv is not thread-safe across a single `gerbv_project_t`. If you process files in parallel, give each thread its own project.
- For untrusted input, sandbox it like any other codec.

If you can test the VS 2022 steps against a current MSYS2 build and report back, that would help us confirm the guide before merging.

PR #484, opened 2026-05-14
docs: add libgerbv consumer guide (Linux/macOS/MSYS2/MSVC)

rampageservices:
## Summary

Adds `doc/example-code/README.md` — a build-and-link guide for downstream
projects that want to use libgerbv. The existing `example1.c`–`example6.c`
files only document the Linux `pkg-config` invocation in their header
comments; there is no top-level entry point for someone arriving with
"how do I call this from my application?".

The new guide covers:

- what libgerbv is, the public header, and a tour of `example1`–`example6`
- a minimal parse + PNG render snippet (the most common headless use case)
- build/link instructions via `pkg-config` on Linux and macOS
- native Windows build via the `msys2-ucrt64-gcc` preset
- consuming the resulting DLL from **Visual Studio 2022 / MSVC** — generating
  an MSVC import library with `gendef` + `lib /def`, the required include
  directories and runtime DLLs, and the `extern "C"` wrapper for C++

Also links the new guide from the "Information for developers" section
of `README.md`.

Addresses #483 (calling libgerbv from VS 2022 / C++ DLL on Windows x64).

## Test plan

- [x] Renders correctly on GitHub (Markdown only — no code changes)
- [x] All links in the new file resolve to in-repo paths
- [ ] Optional follow-up: have a Windows user verify the `gendef` / `lib /def`
      steps against a current MSYS2 UCRT64 build

### DIFF
--- a/README.md
+++ b/README.md
@@ -307,6 +307,9 @@ wishing to incorporate Gerber parsing/editing/exporting/rendering into other
 programs are welcome to use libgerbv. Complete API documentation for libgerbv
 is [here](https://gerbv.github.io/doc/), as well as many example programs using libgerbv.
 
+See [`doc/example-code/README.md`](doc/example-code/README.md) for a guide to
+building against libgerbv on Linux, macOS, MSYS2, and Visual Studio 2022.
+
 <details>
   <summary>Click for Example 1</summary>
    <p>Description: Loads example1-input.gbx into a project, and then exports the layer back to another RS274X file 
--- a/doc/example-code/README.md
+++ b/doc/example-code/README.md
@@ -17,6 +17,11 @@ This directory contains small, runnable examples:
 
 Full HTML API reference: <https://gerbv.github.io/doc/>
 
+Sample Gerber, Excellon drill, and pick-and-place files from a variety of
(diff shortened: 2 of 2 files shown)


rampageservices:
Worth tying this to the two open PRs that close the loop on `doc/example-code/`:

- **#377** (us) — adds `test/CMakeLists.txt` entries that compile `example1`–`example6` and run examples 1–5 under ctest, plus a `test_libgerbv.c` consumer-link target that catches the dxflib symbol-resolution case from #366. Reviewed by @spe-ciellt, all three review points addressed in 1b3f343, currently mergeable=clean.
- **#427** (@eyal0) — drops `continue-on-error: true` from the `Run tests` step so ctest failures actually gate CI. Without it, the new ctest targets from #377 can silently regress.

With #484 + #377 + #427 landed:
- this PR documents how downstream projects consume `libgerbv`
- #377 makes every example program compile and run on every push
- #427 makes a broken example fail CI instead of being swallowed

@spe-ciellt @eyal0 — would be great to get #377 and #427 across the line so this guide stays honest. Happy to rebase any of the three if conflicts appear from the merge order.

PR #490, opened 2026-07-31
Fix two minor compile issues

chrschmidt:
One fix is for the use of strncasecmp() without a proper include in main.c (an alternative fix is described in the commit), the other improves auto-detection for canonicalize_file_name on glibc systems where the user did not specify -D_GNU_SOURCE such that in the case of presence of the function in the library, but not in the header files, the internal declaration is automatically enabled.


spe-ciellt:
Thank you for your PR. Looks good reviewing it. 

Started a PR build and realized that GitHub have changed things during the last month that I need to fix before merging. But I will merge it as soon as I fixed the Windows problem

